PR #39, opened 2023-03-03
Remove git-version gradle plugin

kojandy:
Removes git-version since it is not in use


taeyeon-Kim (approved):

junoyoon:
why? don't we provide the related git version info in "/"?

Issue #52, opened 2023-03-13
Define development guide


junoyoon:
* We didn't define development style yet surprisingly but used precisely defined editorconfig only.
* Following guides are necessary.
  * Coding style 
  * Branch / Merging policy
  * Release policy

Now.. it's the time.

taeyeon-Kim:
I just tried to release 1.0.4 and...  I saw that I hadn't changed the version below... I thought I'd make a decision at this point.
https://github.com/naver/scavenger/blob/develop/build.gradle.kts#L7
```kotlin
allprojects {
    group = "com.navercorp.scavenger"
    version = "1.0.3"
```

junoyoon:
Yes.. it's time to use gradle release plugin like (https://github.com/researchgate/gradle-release)

junoyoon:
what about investigating gradle release plugin? @sohyun-ku

sohyun-ku:
ok. I'm on vacation this week, so I'll check next week.

junoyoon:
OK. Take your time and see u next week. :-)

sohyun-ku:
I think current release step like below

1. make a `release` branch to modify version
2. `release` → `main` PR
3. make a TAG & Release new version at `main` 

If we use `gradle-release plugin`, I think some features make problem.
> • Creates a release tag with the current version.
> • Commits the project with the new version.

First, if we run `gradle release` at release branch, tag is created at release branch.
Second, committed version is `v1.0.1(new version)` but created tag version is `v 1.0.0(current version)`

I think’t the above problems don’t follow `3. make a TAG & Release new version at main`

Why don’t we use Github Actions instead of gradle-release-plugin?
Using a github action like below we can fire an event(ex. release, tag) when the `main` branch is pushed.

- https://github.com/actions/create-release
- https://github.com/actions/upload-release-asset

If I've missed anything, please tell me.

sohyun-ku:
@junoyoon @taeyeon-Kim @kojandy 
https://github.com/naver/scavenger/issues/52#issuecomment-1477392705 What about this?

kojandy:
I also think github actions might be a better solution as it can generate and upload all the release assets without manual intervention.

junoyoon:
I think we can create release branch from master to activate the github action to release.
and no merge back this branch to master. just drop it.

what about this scheme?

sohyun-ku:
Sounds good :)
Then I'll add github actions when pushed release branch.

sohyun-ku:
[...]

PR #55, opened 2023-03-28
Add release github actions

sohyun-ku:
#52 
~~When pushed to remote repository`release/*` branch, create release using github actions.~~
~~The behavior is as follows.~~
~~1. Checkout code~~
~~2. Setup JDK~~
~~3. Build project~~
~~4. Extract version from commit message~~
    ~~- Commit messages should include the release version.~~
~~5. Create Release~~

plz refer https://github.com/naver/scavenger/pull/55#issuecomment-1489993218

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,8 +1,12 @@
 name: Release
+
+run-name: Create Release
+
 on:
   push:
-    branches:
-      - 'release/*'
+    tags:
+      - "v*.*.*"
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


kojandy (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Setup Java JDK
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: '11'
+      - name: Build project
+        run: ./gradlew assemble
```
junoyoon, line 18: This only do assemble...
Don't we need release plugin replace SNAHOP version to empty string?

taeyeon-Kim: `assemble` doesn't run a test task. How about replacing it with `build` task?

sohyun-ku: @junoyoon 
This action is just create release & tag when pushed `release/*` branch.
So if we want to replace SNAPSHOT version to empty string, we'll have to do it by self.

I was some confusion because the operation of the `gradle-release plugin` was different from the current release method.
It seems to have a similar operation to what you said this https://github.com/naver/scavenger/issues/52#issuecomment-1485345568, so I'll check again

sohyun-ku: @junoyoon 

We can use `gradle-release` plugin **if merging `release` branch to `main` is ok.**
Using `gradle-release` commits tag version and new version at working branch. 
We want to create tag at release branch, we can use `pushReleaseVersionBranch ` option. ([reference](https://github.com/researchgate/gradle-release#introduction))
But using `pushReleaseVersionBranch` will checkout to the release branch and merge from the working branch.
If we don't use `pushReleaseVersionBarnch` option, just commit working branch.(in our case maybe working branch is `main`)

If we decide to use `gradle-release` plugin, I'd like to proceed release step as follows.
```
1. Checkout release branch and commit release version by developer.
  - It can be replaced with the gradle release plugin, but for some reason it is difficult to control it with the github action.
2. Using github actions on push event at release branch, trigger gradle-release plugin. 
  - This step is creating TAG.
3. Using github actions on tag event, create Release.
```

cc. @naver/scavenger-dev

sohyun-ku:
@junoyoon @naver/scavenger-dev 

If merging release branch to main is ok, 
What about this?

release step as follows.
```
1. develop -> main PR
  - If we want to release, PR title should be include release ${version}
2. If PR is merged, trigger tag github action
  - Remove SNAPSHOT and commit to next version by gradle-release plugin
  - Create Tag in this action
3. If a tag is created, trigger create a release github action
  - Create Release in this action
```

and git log is as follows.
image

junoyoon:
very good!!! go for it!!

taeyeon-Kim (approved):

sohyun-ku:
I rebased it with the latest `develop` branch.

taeyeon-Kim (approved):

.github/workflows/release.yml
junoyoon: i will review this weekend? ok?

sohyun-ku: @junoyoon 
https://github.com/naver/scavenger/pull/55#discussion_r1151478029 is outdated.
plz refer https://github.com/naver/scavenger/pull/55#issuecomment-1489993218
When release, we only have to Pull Request `develop` to `main` branch.
Other tasks are done as Github Actions.

Issue #69, opened 2023-05-02
Use vue-json-csv instead of kotlin-csv


taeyeon-Kim:
Use `vue-json-csv` instead of `kotlin-csv`
- vue-json-csv: https://github.com/Belphemur/vue-json-csv

it has the advantage of not handling IO on the server and doesn't require direct control of FE's DOM.

taeyeon-Kim:
😢

PR #70, opened 2023-05-08
Use xlsx instead of kotlin-csv

sohyun-ku:
#69 

I used `xlsx` instead of `vue-json-csv`. Becuz I want to create a csv file after fetching the snapshot data using callback.


scavenger-api/src/main/kotlin/com/navercorp/scavenger/controller/SnapshotController.kt
```diff
@@ -105,31 +97,12 @@ class SnapshotController(
         )
     }
 
-    @GetMapping("/customers/{customerId}/snapshot/{snapshotId}/export", produces = ["text/csv"])
+    @GetMapping("/customers/{customerId}/snapshot/{snapshotId}/export")
```
taeyeon-Kim, line 100:
```suggestion
    @GetMapping("/customers/{customerId}/snapshots/{snapshotId}/export")
```

taeyeon-Kim:
We don't use `kotlin-csv` anymore, so please delete it as well.

scavenger-frontend/src/components/snapshot/SnapshotTable.vue
```diff
@@ -85,6 +85,8 @@ import SnapshotForm from "./SnapshotForm.vue";
 import Momnet from "moment";
 import {useStore} from "../util/store";
 import {ElMessageBox, ElNotification} from "element-plus";
+import * as Xlsx from 'xlsx'
```
taeyeon-Kim, line 88: Previously, the delimiter was tab. Is this the same in xsx?

sohyun-ku: In xlsx, the default field separator delimiter is `,`

sohyun-ku:
@taeyeon-Kim 

> Is the reason we can't use vue-json-csv is because we can't handle the download event on click event after calling the api?

If we use vue-json-csv, we can use like below.
```
<download-csv
    :data   = "json_data">
    Download Data
    <img src="download_icon.png">
</download-csv>
```
But...There was a problem in the process of receiving data asynchronously when a click event occurred. 
There may be several workarounds, but none of the features are explicitly provided by the package. (https://github.com/Belphemur/vue-json-csv/issues/175)
The reason why I used xlsx is that I wanted to implement a function that asynchronously call data when a click event occurs and creates an excel file with a callback.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/controller/SnapshotController.kt
```diff
@@ -105,31 +97,12 @@ class SnapshotController(
         )
     }
 
-    @GetMapping("/customers/{customerId}/snapshot/{snapshotId}/export", produces = ["text/csv"])
+    @GetMapping("/customers/{customerId}/snapshots/{snapshotId}/export")
```
taeyeon-Kim, line 100: Should we also rename the methods across the board?
It seems like it's now an API to get all snapshot nodes rather than export.

sohyun-ku:
There is no body size limit for HTTP responses, but added paging because it can take a long time to parse large size JSON data in the browser.
The execution time based on 10000 rows in the local env is within 1s.

taeyeon-Kim (approved):

scavenger-frontend/package.json
```diff
@@ -30,7 +30,9 @@
     "splitpanes": "^3.1.5",
     "vue": "^3.2.41",
     "vue-i18n": "^9.2.2",
-    "vue-router": "^4.1.6"
+    "vue-loading-overlay": "^6.0.3",
```
taeyeon-Kim, line 33: 👍

taeyeon-Kim (approved):

taeyeon-Kim (approved):

scavenger-api/src/main/kotlin/com/navercorp/scavenger/repository/SnapshotNodeDao.kt
```diff
@@ -5,28 +5,46 @@ import com.navercorp.scavenger.entity.SnapshotNodeEntity
...
 @Repository
 class SnapshotNodeDao(
     entityJdbcProvider: EntityJdbcProvider,
     snapshotNodeRepository: SnapshotNodeRepository
-) : JdbcDaoSupport(entityJdbcProvider), SnapshotNodeRepository by snapshotNodeRepository {
+) : JdbcDaoSupport(entityJdbcProvider), SingleValueSelectTrait, SnapshotNodeRepository by snapshotNodeRepository {
     private val sql: SnapshotNodeSql = super.sqls(::SnapshotNodeSql)
 
-    fun findAllExportSnapshotNode(
+    fun selectAllExportSnapshotNode(
```
junoyoon, line 18: we use find here.

scavenger-frontend/src/components/snapshot/SnapshotTable.vue
```diff
@@ -165,23 +166,31 @@ export default {
...
+          const data = response.data.map(row => {
+            row.filterInvokedAtMillis = Moment.unix(row.filterInvokedAtMillis).format("YYYY.MM.DD HH:mm:ss");
+            row.lastInvokedAtMillis = row.lastInvokedAtMillis == null ? null : Moment.unix(row.lastInvokedAtMillis).format("YYYY.MM.DD HH:mm:ss");
+            return row;
+          });
+          utils.sheet_add_json(workSheet, data, {header, origin: -1, skipHeader: true})
+        }
+
+        utils.book_append_sheet(workBook, workSheet);
+        writeFile(workBook, `snapshot_${id}.csv`);
```
junoyoon, line 188: @sohyun-ku 

Could u explain why xlsx should be created in fe layer? what was your intention here?

taeyeon-Kim: I opened this issue #69.
In my opinion... it has the advantage of not handling IO on the server and doesn't require direct control of FE's DOM.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/repository/SnapshotNodeDao.kt
sohyun-ku: fixed in 90c8f8a39bdbfcbafd0070c2c7150a5cc9821932

scavenger-frontend/src/components/snapshot/SnapshotTable.vue
junoyoon: I don't think this is really necessary.
This fix blocks the way directly download excel using curl. Moreover, it always need to use the browser.
Instead of this... what about adding excel download beside of csv?

sohyun-ku: I hadn't considered using curl too..haha..

> what about adding excel download beside of csv?

Before I did this PR, there was already excel download feature, what does this mean?

taeyeon-Kim: @junoyoon Please check out sohyun`s comment.

junoyoon: @sohyun-ku @taeyeon-Kim 
Sorry for late response. I was too busy.. :-(

It was csv export feature not excel.
And csv is created in server side.
So it's callable using cli

I think we should make it server side job (not fe).

sohyun-ku: @junoyoon @taeyeon-Kim 

thanks for explaining :)
I didn't think someone can use cli when using csv export feature...
We have already a csv export feature on the server side before this PR.
So I'll close this PR.

PR #73, opened 2023-05-17
Add enter event when create workspace

sohyun-ku:
Add `enter` event to input field and empty string validation when create workspace.


taeyeon-Kim:
For validation events, see `ConfForm.vue`.
We are using validation from element-plus.
https://element-plus.org/en-US/component/form.html#validation

sohyun-ku:
@taeyeon-Kim 

thx:)
amend b1b6649499707e5544397532c0fba817dee9a984

image

taeyeon-Kim (approved):
👍

Issue #75, opened 2023-05-17
GC not working properly


taeyeon-Kim:
Looking at the `agent_state` table, it looks like GC is not working properly.

PR #76, opened 2023-05-17
Add sweepAgentStatesAndJvms chunk

taeyeon-Kim:
resolved: #75

### DIFF
--- a/scavenger-collector/src/main/kotlin/com/navercorp/scavenger/repository/sql/AgentStateSql.kt
+++ b/scavenger-collector/src/main/kotlin/com/navercorp/scavenger/repository/sql/AgentStateSql.kt
@@ -26,6 +26,7 @@ class AgentStateSql : SqlGeneratorSupport() {
         WHERE
             customerId = :customerId
             AND lastPolledAt < :lastPolledAt
+        LIMIT 100000
         """.trimIndent()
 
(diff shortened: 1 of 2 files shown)


kojandy (approved):

scavenger-collector/src/main/kotlin/com/navercorp/scavenger/repository/sql/AgentStateSql.kt
```diff
@@ -26,6 +26,7 @@ class AgentStateSql : SqlGeneratorSupport() {
         WHERE
             customerId = :customerId
             AND lastPolledAt < :lastPolledAt
+        LIMIT 100000
```
junoyoon, line 29: isn't it little big? what about use 10000

junoyoon (commented):
commented

scavenger-collector/src/main/kotlin/com/navercorp/scavenger/repository/sql/AgentStateSql.kt
taeyeon-Kim: because the row count limit in vitess is 100000.I changed it to 10000.

junoyoon (approved):
LGTM
Merge Plz

PR #79, opened 2023-06-30
sweep jvms without agent

tngusdl3719:
Hello,

I'm currently using a scavenger and I've encountered a problem where deleted methods are not getting garbage collected. Upon inspecting the collector code, it seems that the method garbage collection marks methods collected before the earliest CreatedAt value of CodebaseFingerprints owned by the Customer, then proceeds with the sweep. Also, CodebaseFingerprints are updated along with the JVM.

For JVM to be updated, an AgentState is required. However, there was a JVM that didn't have an AgentState. I'm unsure of the cause of the non-existent AgentState, but it seems that if a JVM doesn't have an AgentState, it would forever remain in the DB.

So, I've created a method to sweep the JVM without an AgentState.

Please review the PR and provide your feedback.

### DIFF
--- a/scavenger-collector/src/main/kotlin/com/navercorp/scavenger/repository/sql/JvmSql.kt
+++ b/scavenger-collector/src/main/kotlin/com/navercorp/scavenger/repository/sql/JvmSql.kt
@@ -56,19 +56,6 @@ class JvmSql : SqlGeneratorSupport() {
             AND uuid IN ( :uuids )
         """.trimIndent()
 
-    fun deleteAllByWithoutAgent(): String =
-        """
-            DELETE FROM
-                jvms
-            WHERE
-                uuid NOT IN (
-                    SELECT
-                        jvmUuid
-                    FROM
-                        agent_state
-                )
-        """.trimIndent()
-
     fun selectUuidsByWithoutAgent(): String =
         """
            SELECT uuid FROM
--- a/scavenger-collector/src/main/kotlin/com/navercorp/scavenger/repository/sql/JvmSql.kt
+++ b/scavenger-collector/src/main/kotlin/com/navercorp/scavenger/repository/sql/JvmSql.kt
@@ -55,4 +55,17 @@ class JvmSql : SqlGeneratorSupport() {
             customerId = :customerId
             AND uuid IN ( :uuids )
         """.trimIndent()
+
+    fun deleteAllByWithoutAgent(): String =
+        """
+            DELETE FROM
+                jvms
+            WHERE
+                uuid NOT IN (
+                    SELECT
+                        jvmUuid
+                    FROM
+                        agent_state
+                )
+        """.trimIndent()
 }
--- a/scavenger-collector/src/main/kotlin/com/navercorp/scavenger/repository/sql/JvmSql.kt
+++ b/scavenger-collector/src/main/kotlin/com/navercorp/scavenger/repository/sql/JvmSql.kt
@@ -56,7 +56,7 @@ class JvmSql : SqlGeneratorSupport() {
             AND uuid IN ( :uuids )
         """.trimIndent()
 
-    fun selectUuidsByWithoutAgent(): String =
+    fun selectAllUuidsByWithoutAgent(): String =
         """
            SELECT uuid FROM
                 jvms
--- a/scavenger-collector/src/main/kotlin/com/navercorp/scavenger/repository/sql/JvmSql.kt
+++ b/scavenger-collector/src/main/kotlin/com/navercorp/scavenger/repository/sql/JvmSql.kt
@@ -68,4 +68,18 @@ class JvmSql : SqlGeneratorSupport() {
                         agent_state
                 )
         """.trimIndent()
+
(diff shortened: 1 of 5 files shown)


scavenger-collector/src/main/kotlin/com/navercorp/scavenger/repository/sql/JvmSql.kt
```diff
@@ -55,4 +55,17 @@ class JvmSql : SqlGeneratorSupport() {
             customerId = :customerId
             AND uuid IN ( :uuids )
         """.trimIndent()
+
+    fun deleteAllByWithoutAgent(): String =
```
taeyeon-Kim, line 59: @tngusdl3719 Scavenger is also supported by Vitess. Vitess does not support subqueries in sharded DML.
Also, we're using a shard key set to customerId.
How do you like these ways?

https://github.com/naver/scavenger/blob/d46ff3eb5d876e973736855e13a6f23d3acf03c7/scavenger-collector/src/main/kotlin/com/navercorp/scavenger/service/GarbageCollectService.kt#L97-L113


Also, we're using a shard key set to customerId.
How do you like these ways?
Before executing 'jvmDao.deleteAllByCustomerIdAndUuids', execute the following query.

- query the JVM, which does not exist in the agent with a shard key.
```sql
SELECT uuid FROM
    jvms
WHERE
    customerId = :customerId
    AND uuid NOT IN (
        SELECT
            jvmUuid
        FROM
            agent_state
    );
```

execute `jvmDao.deleteAllByCustomerIdAndUuids` with the result of the query.
```kotlin
jvmDao.deleteAllByCustomerIdAndUuids(customerId, agentStates.map { it.jvmUuid } + results)
```

taeyeon-Kim (commented):
Firstly, thank you so much for reporting these issues.
We are in the process of checking why this is occurring.

scavenger-collector/src/main/kotlin/com/navercorp/scavenger/repository/sql/JvmSql.kt
tngusdl3719: I overlooked some aspects of Vitess and sharding.

I have made changes based on the feedback you provided.

thx

scavenger-collector/src/main/kotlin/com/navercorp/scavenger/service/GarbageCollectService.kt
```diff
@@ -101,8 +101,9 @@ class GarbageCollectService(
                     customerId,
                     baseDateTime.minusMillis(intervalService.batchSweepMarginMilliSecond)
                 )
+                var jvmUuids = jvmDao.findUuidsByWithoutAgent(customerId)
```
taeyeon-Kim, line 104:
```suggestion
                val jvmUuids = jvmDao.findUuidsByWithoutAgent(customerId)
```

scavenger-collector/src/main/kotlin/com/navercorp/scavenger/service/GarbageCollectService.kt
```diff
@@ -101,8 +101,9 @@ class GarbageCollectService(
                     customerId,
                     baseDateTime.minusMillis(intervalService.batchSweepMarginMilliSecond)
                 )
+                val jvmUuids = jvmDao.findAllUuidsByWithoutAgent(customerId)
```
sohyun-ku, line 104: How about `uuidsWithoutAgent` instead of `jvmUuids`?

tngusdl3719: Doesn't a method name that explicitly states "find UUIDs without an agentState" seem clearer?

tngusdl3719: > How about `uuidsWithoutAgent` instead of `jvmUuids`?

Doesn't a method name that explicitly states "find UUIDs without an agentState" seem clearer?

scavenger-collector/src/main/kotlin/com/navercorp/scavenger/service/GarbageCollectService.kt
```diff
@@ -101,8 +101,9 @@ class GarbageCollectService(
                     customerId,
                     baseDateTime.minusMillis(intervalService.batchSweepMarginMilliSecond)
                 )
+                val uuidsWithougAgent = jvmDao.findAllUuidsByWithoutAgent(customerId)
 
-                jvmDao.deleteAllByCustomerIdAndUuids(customerId, agentStates.map { it.jvmUuid })
+                jvmDao.deleteAllByCustomerIdAndUuids(customerId, agentStates.map { it.jvmUuid } + uuidsWithougAgent)
```
sohyun-ku, line 106:
```suggestion
                jvmDao.deleteAllByCustomerIdAndUuids(customerId, agentStates.map { it.jvmUuid } + uuidsWithoutAgent)
```

taeyeon-Kim (approved):

sohyun-ku (approved):

PR #83, opened 2023-07-06
Add chunk to selectAllUuidsByWithoutAgent

sohyun-ku:
related #76

### DIFF
--- a/scavenger-collector/src/main/kotlin/com/navercorp/scavenger/service/GarbageCollectService.kt
+++ b/scavenger-collector/src/main/kotlin/com/navercorp/scavenger/service/GarbageCollectService.kt
@@ -101,13 +101,18 @@ class GarbageCollectService(
                     customerId,
                     baseDateTime.minusMillis(intervalService.batchSweepMarginMilliSecond)
                 )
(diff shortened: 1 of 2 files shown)


scavenger-collector/src/main/kotlin/com/navercorp/scavenger/service/GarbageCollectService.kt
```diff
@@ -101,9 +101,14 @@ class GarbageCollectService(
                     customerId,
                     baseDateTime.minusMillis(intervalService.batchSweepMarginMilliSecond)
                 )
-                val uuidsWithoutAgent = jvmDao.findAllUuidsByWithoutAgent(customerId)
 
-                jvmDao.deleteAllByCustomerIdAndUuids(customerId, agentStates.map { it.jvmUuid } + uuidsWithoutAgent)
+                do {
```
taeyeon-Kim, line 105: Expect a lot of loops.
Instead of a nested while, how about a separate while?

sohyun-ku: fixed in e0d780ac0e93dc2d4a9bb464888a4ae25199477d

taeyeon-Kim (approved):
👍

PR #87, opened 2023-08-18
Use Gradle JVM toolchain and exclude old agent from building

kojandy:
- Gradle handles required JDK version automatically regardless of the local Java version.
- Excludes the building of `scavenger-old-agent-java` when `./gradlew build` is run. `-PoldAgent` option can be used to re-include the subproject.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -27,8 +27,8 @@ release {
         version
     }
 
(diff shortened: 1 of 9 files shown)


sohyun-ku (approved):

taeyeon-Kim (approved):
👍

Issue #93, opened 2023-09-12
UI : Package Naming Case Sensitivity Issue


hyukjin-lee:
Hello,

We are currently using Scavenger in our legacy system. There are projects where package names are started with uppercase. It appears that on the Scavenger UI, the starting character's case (uppercase or lowercase) is used to distinguish between a package and a class.

Is there any possibility to improve this so that Scavenger only recognizes actual packages as packages, irrespective of the case of the starting character? This would help in ensuring more accurate identification and reduce confusion.

Thank you for your attention to this matter.

sohyun-ku:
@hyukjin-lee 

Hi :)
This feature was implemented in #96.
It'll be included in release `v1.1.2`.

PR #96, opened 2023-09-19
Implement package name case sensitive

sohyun-ku:
#93 

The standard that seperate `method`, `class`, `package` as follows
```
1. if contains `(`, it's method.
  - Because the signature is taken from methodInvocationTable. 
  - I think always last element is method.
2. if contains `$`, it's (inner) class
3. if next element is `(`(method) or `$`(inner class), it is class.
4. Otherwise, it's package
```

image

### DIFF
--- a/scavenger-api/src/main/kotlin/com/navercorp/scavenger/service/SnapshotNodeService.kt
+++ b/scavenger-api/src/main/kotlin/com/navercorp/scavenger/service/SnapshotNodeService.kt
@@ -129,11 +129,11 @@ class SnapshotNodeService(
         val lastElement = elements.last()
         lastElement["signature"] = "${lastElement["signature"]}($arguments" + if (arguments.last() != ')') ")" else ""
 
-        elements.asReversed().forEachIndexed { index, mutableMap ->
+        elements.forEachIndexed { index, mutableMap ->
             val signature = checkNotNull(mutableMap["signature"]) { "signature must be not null" }
             mutableMap["type"] = if (signature.contains("(")) {
                 Node.Type.METHOD.name
-            } else if (signature.contains("$") || checkNotNull(elements[elements.size - index]["signature"]).contains("[($]".toRegex())) {
+            } else if (signature.contains("$") || checkNotNull(elements[index + 1]["signature"]).contains("[($]".toRegex())) {
                 Node.Type.CLASS.name
             } else {
                 Node.Type.PACKAGE.name
--- a/scavenger-api/src/main/kotlin/com/navercorp/scavenger/service/SnapshotNodeService.kt
+++ b/scavenger-api/src/main/kotlin/com/navercorp/scavenger/service/SnapshotNodeService.kt
@@ -24,7 +24,7 @@ class SnapshotNodeService(
         filteredMethodInvocations.forEach {
             updateCountGraph(
                 current = root,
-                elements = splitSignature(it),
+                elements = splitSignatureWithType(it),
                 isUsed = it.invokedAtMillis > 0 && it.invokedAtMillis >= snapshotEntity.filterInvokedAtMillis,
                 invokedAtMillis = it.invokedAtMillis
             )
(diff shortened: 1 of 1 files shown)


scavenger-api/src/main/kotlin/com/navercorp/scavenger/service/SnapshotNodeService.kt
```diff
@@ -112,40 +113,40 @@ class SnapshotNodeService(
...
             .toMutableList()
 
         if (isConstructor(methodInvocationEntity)) {
-            elements.add(elements.last())
+            val constructorSignature = checkNotNull(elements.last()["signature"]) { "signature must be not null" }
+            elements.add(mutableMapOf("signature" to constructorSignature.replace("$", "")))
+        }
+
+        val lastElement = elements.last()
+        lastElement["signature"] = "${lastElement["signature"]}($arguments" + if (arguments.last() != ')') ")" else ""
```
taeyeon-Kim, line 130: Is any case other than `)` at the end of the argument?

scavenger-api/src/main/kotlin/com/navercorp/scavenger/service/SnapshotNodeService.kt
```diff
@@ -112,40 +113,40 @@ class SnapshotNodeService(
...
+
+        elements.asReversed().forEachIndexed { index, mutableMap ->
+            val signature = checkNotNull(mutableMap["signature"]) { "signature must be not null" }
+            mutableMap["type"] = if (signature.contains("(")) {
+                Node.Type.METHOD.name
+            } else if (signature.contains("$") || checkNotNull(elements[elements.size - index]["signature"]).contains("[($]".toRegex())) {
```
taeyeon-Kim, line 136: How about removing `asReversed()` and just doing `elements(index + 1)`?

scavenger-api/src/main/kotlin/com/navercorp/scavenger/service/SnapshotNodeService.kt
```diff
@@ -75,22 +75,23 @@ class SnapshotNodeService(
 
     private fun updateCountGraph(
         current: Node,
-        elements: List<String>,
+        elements: List<Map<String, String>>,
```
taeyeon-Kim, line 78: How about writing a data class instead of a map?

scavenger-api/src/main/kotlin/com/navercorp/scavenger/service/SnapshotNodeService.kt
sohyun-ku: The reason I used as `Reversed()` was that I thought we had to go in reverse order to distinguish between `package` and `classe`.
Before, if the first letter is a capital letter or parent type is a `class`, we thought it was a `class`.
Now we can't use the first one, so I don't think there's a way to distinguish between `package` and `class` if we go in order.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/service/SnapshotNodeService.kt
sohyun-ku: Good. I had to check the safety when I used the map.
If we use a data class, we don't have to check it

scavenger-api/src/main/kotlin/com/navercorp/scavenger/service/SnapshotNodeService.kt
sohyun-ku: maybe as follow case.
if end like this ...(truncated), we add `)` manually.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/service/SnapshotNodeService.kt
taeyeon-Kim: Yes, you're right, I thought that might be something that could be checked at `elements[index +1]["signature"]).contains("[($]".toRegex())`.

sohyun-ku: um....
if we use `asReversed()` and `forEachIndexed` together, the index is not in reverse order. It starts from 0....

sohyun-ku: I had missed it. If we search in order as you said, we can distinguish between `package` and `class`.
I'll fix it.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/service/SnapshotNodeService.kt
sohyun-ku: fixed in b2d0a4aefe75de4a306302006bc9df595635eac3

scavenger-api/src/main/kotlin/com/navercorp/scavenger/service/SnapshotNodeService.kt
```diff
@@ -179,6 +181,12 @@ class SnapshotNodeService(
         )
     }
 
+    data class SignatureWithType(
+        val signature: String,
+    ) {
+        lateinit var type: Node.Type
```
taeyeon-Kim, line 187: 👍

PR #98, opened 2023-09-22
Modify snapshot_nodes table's signature type to text

sohyun-ku:
#95 
Change `snapshot_nodes` table's `signature` column type to `text`.
And add test data
image

### DIFF
--- a/scavenger-schema/src/main/resources/db/changelog/db.changelog-master.yaml
+++ b/scavenger-schema/src/main/resources/db/changelog/db.changelog-master.yaml
@@ -2,5 +2,7 @@ databaseChangeLog:
   - include:
       file: db/changelog/schema.sql
   - include:
+      file: db/changelog/schema-1.1.2.sql
+  - include:
       file: db/changelog/test-data-set.sql
       context: local
--- a/scavenger-schema/src/main/resources/db/changelog/schema-1.1.2.sql
+++ b/scavenger-schema/src/main/resources/db/changelog/schema-1.1.2.sql
@@ -1,0 +1,3 @@
+--changeset scavenger:3
+
+ALTER TABLE snapshot_nodes MODIFY signature TEXT;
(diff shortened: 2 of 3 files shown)


taeyeon-Kim (commented):
There are two ways to consider this
The first is to append the changeset.
The second is to write a new SQL file and add it to the changelog-master.
https://github.com/naver/scavenger/blob/b3cd7f9b5e4088155929d876e74d4270d9b6d0de/scavenger-schema/src/main/resources/db/changelog/db.changelog-master.yaml#L1-L6



The second way seems easier to manage, what do you think?

scavenger-schema/src/main/resources/db/changelog/test-data-set.sql
```diff
@@ -100,3 +100,222 @@ VALUES (1, 1, 'com.example.demo.additional.AdditionalService.get()', 0, 1,
         'METHOD'),
        (29, 2, 'com.example.demo.MyController', 2, 1, 'com.example.demo', 1, 1646027014967, 'CLASS'),
        (30, 2, 'com.example.demo', 2, 3, '', 1, 1646027014967, 'PACKAGE');
+
+
+--changeset scavenger:4
+
+TRUNCATE TABLE codebase_fingerprints;
```
taeyeon-Kim, line 107: Is it possible to add without truncating?

sohyun-ku:
@taeyeon-Kim 
I think so too, cause we can see changes without having to check the sql file directly.
I'll change it this way :)

scavenger-schema/src/main/resources/db/changelog/test-data-set.sql
sohyun-ku: I think it's possible.
Even though it's still like that...
The test code needs more changes than I thought, so I'll roll back and work on a different PR.

scavenger-schema/src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -1,6 +1,8 @@
 databaseChangeLog:
   - include:
       file: db/changelog/schema.sql
+  - include:
+      file: db/changelog/schema-1.1.1.sql
```
taeyeon-Kim, line 5: maybe... `1.1.2`

sohyun-ku: haha..... fixed in 51ccd12425738493b11b9706a412fdcce621657b

taeyeon-Kim (approved):
👍

PR #101, opened 2023-11-14
Support JDK 21

sohyun-ku:
resolved: #97 

- ASM v9.5 supports to JDK 21 
  - ASM Versions: https://asm.ow2.io/versions.html
- ByteBuddy v1.14.3 uses ASM v9.5
  - https://github.com/raphw/byte-buddy/compare/byte-buddy-1.14.2...byte-buddy-1.14.3

image

PR #102, opened 2023-11-14
Fix implicit dependencies between tasks

sohyun-ku:
Integration Test was disabled when run scavenger-agent-java:build that triggered by PR builder.
To solove this problem, we can use 'mustRunAfter' to specify the order between 'jar' and 'integrationTest' tasks.

https://github.com/naver/scavenger/actions/runs/6860169774/job/18653556996#step:5:102
> Execution optimizations have been disabled for task ':scavenger-agent-java:integrationTest' to ensure correctness due to the following reasons:
>  - Gradle detected a problem with the following location: '/home/runner/work/scavenger/scavenger/scavenger-agent-java/build/libs/scavenger-agent-java-1.1.1-SNAPSHOT.jar'. Reason: Task ':scavenger-agent-java:integrationTest' uses this output of task ':scavenger-agent-java:jar' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.6.1/userguide/validation_problems.html#implicit_dependency for more details about this problem.

### DIFF
--- a/scavenger-agent-java/build.gradle.kts
+++ b/scavenger-agent-java/build.gradle.kts
@@ -98,6 +98,7 @@ val integrationTestRuntimeClasspath = configurations.named("integrationTestRunti
 
 tasks.named<Test>("integrationTest") {
     dependsOn(tasks.shadowJar)
+    mustRunAfter(tasks.jar)
     shouldRunAfter(tasks.test)
     useJUnitPlatform()
 


taeyeon-Kim (approved):
I think there was an issue with enabling jar tasks in #91.
Thanks you 😄 , it would have remained a serious bug if you hadn't checked the result of the action.

taeyeon-Kim:
Since the integration test passed in this PR, can we assume that the `Support JDK 21` #101 PR is okay?

sohyun-ku:
Of course, this PR is after e944d3825c2475bdb2aaacd5540afac2e44921c6 commit :)

kojandy (approved):

Issue #103, opened 2023-11-17
Search can't find classes if package name contains numbers


taeyeon-Kim:
When I search for a class under a package that contains a number, such as "com.example.demo.service2.TestService", it fails to navigate to the class.
<img width="296" alt="image" src="https://github.com/naver/scavenger/assets/27491403/035aa2a7-20df-4d29-bcb3-8ed38653db93">

<img width="315" alt="image" src="https://github.com/naver/scavenger/assets/27491403/095bc7ba-cd77-42bc-b66d-1d384ea7cd82">

PR #104, opened 2023-11-17
Fix to navigate signatures when package names start with a single prefix

taeyeon-Kim:
resolved: #103

### DIFF
--- a/run-agent.sh
+++ b/run-agent.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-./gradlew build -p scavenger-agent-java -x test
+./gradlew build -p scavenger-agent-java -x test -x integrationTest
 ./gradlew build -p scavenger-demo -x test
 
 java -Dscavenger.configuration=./scavenger-demo/scavenger.conf  \
--- a/scavenger-frontend/src/components/snapshot/SnapshotDetail.vue
+++ b/scavenger-frontend/src/components/snapshot/SnapshotDetail.vue
@@ -182,7 +182,7 @@ export default {
           signature = candidate.signature;
         } else {
           signature = this.snapshotData.children
-            .find(child => querySignature.startsWith(child.signature)).signature;
+            .find(child => this.isStartsWithSignature(querySignature, child.signature)).signature;
         }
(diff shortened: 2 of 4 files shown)


run-agent.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-./gradlew build -p scavenger-agent-java -x test
+./gradlew build -p scavenger-agent-java -x test -x integrationTest
```
taeyeon-Kim, line 3: integrationTest was slowing down the build, so it was excluded.

scavenger-frontend/src/components/snapshot/SnapshotDetail.vue
```diff
@@ -328,6 +328,17 @@ export default {
...
+    isStartsWithSignature(querySignature, signature) {
+      if (!querySignature.startsWith(signature)) {
+        return false;
+      }
+
+      if (querySignature === signature) {
+        return true;
+      }
+
+      return querySignature.substring(signature.length).startsWith(".");
```
sohyun-ku, line 340: What is this line means?

scavenger-frontend/src/components/snapshot/SnapshotDetail.vue
```diff
@@ -328,6 +328,17 @@ export default {
     resizeHorizontal(size) {
       localStorage.setItem("scavenger.snapshot.horizontal-size", size);
     },
+    isStartsWithSignature(querySignature, signature) {
+      if (!querySignature.startsWith(signature)) {
+        return false;
+      }
+
+      if (querySignature === signature) {
```
taeyeon-Kim, line 336: This condition is true when searching for a class.
```kotlin
      if (querySignature === signature) {
        return true;
      }
```

scavenger-frontend/src/components/snapshot/SnapshotDetail.vue
taeyeon-Kim: The package must be followed by a class, so check for a `.` as well.

sohyun-ku (approved):

kojandy (approved):

PR #107, opened 2023-11-24
Add foojay resolver plugin

taeyeon-Kim:
Add the foojay-resolver plugin to automatically install the JDK if it is not installed at build time.
https://docs.gradle.org/current/userguide/toolchains.html#sub:download_repositories

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -13,3 +13,7 @@ include("scavenger-old-model")
 include("scavenger-demo")
 include("scavenger-demo-extension")
 include("scavenger-frontend")
+
+plugins {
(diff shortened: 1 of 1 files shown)


sohyun-ku:
Do we need `foojay-resolver` even though we're using `toolchain`?

> By default, Gradle automatically detects local JRE/JDK installations so no further configuration is required by the user. 
> https://docs.gradle.org/current/userguide/toolchains.html#sec:auto_detection

taeyeon-Kim:
@sohyun-ku If Gradle can’t find a locally available toolchain that matches the requirements of the build, it can automatically download one.

sohyun-ku (approved):

kojandy (approved):

PR #116, opened 2024-01-31
Modify extract version regexp in release workflow

sohyun-ku:
The patch release has also been changed to enable github actions.

image


kojandy (approved):

taeyeon-Kim (approved):
👍

PR #118, opened 2024-02-20
Modify snapshot_nodes table's signature column to not null

sohyun-ku:
related to #98

If the ddl script does not contain 'NOT NULL', it will change to a nullable column.
Reset the signature column to the not null field.


scavenger-schema/src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -2,7 +2,7 @@ databaseChangeLog:
   - include:
       file: db/changelog/schema.sql
   - include:
-      file: db/changelog/schema-1.1.2.sql
+      file: db/changelog/schema-1.1.3.sql
```
taeyeon-Kim, line 5: How about adding `include` instead of fixing it?

```yaml
 - include:
      file: db/changelog/schema-1.1.2.sql
 - include:
      file: db/changelog/schema-1.1.3.sql
```

sohyun-ku: fixed in https://github.com/naver/scavenger/pull/118/commits/42e9b59b40ce8c12e39bafd45a8425de4f3bd961

taeyeon-Kim (approved):
👍

Issue #122, opened 2024-05-16
Bump up Spring Boot 3


sohyun-ku:
Bump up Spring Boot 3

sohyun-ku:
Job completed.
I'll be included in the next release.

PR #124, opened 2024-05-28
Bump up spring boot 3

sohyun-ku:
related to #122 

- springBoot 2.5.12 -> 3.2.4
- springDependencyManagement 1.0.11 -> 1.1.4
- java 11 -> 17
- kotlin 1.8.10 -> 1.9.0
- `mysql:mysql-connector-java` -> `com.mysql:mysql-connector-j`
  - https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#mysql-jdbc-driver
- `javax.servlet` → `jakarta.servlet`
  - https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#jakarta-ee
- Modify `EntityConvertibleSqlParameterSourceFactory` signature
  - https://github.com/naver/spring-jdbc-plus/compare/3.0.1...3.2.4#diff-44f263b68bf4136ef03fae9d1595a5eb4528cb238a1869bb7be2e5edb5943996
- `verifyZeroInteractions` → `verifyNoInteractions`Deprecated `verifyZeroInteractions`
  - Deprecated `verifyZeroInteractions`
  - https://www.javadoc.io/doc/org.mockito/mockito-core/3.11.1/org/mockito/Mockito.html#verifyZeroInteractions-java.lang.Object
- Remove `@ConstructorBinding`
  - https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0.0-M2-Release-Notes#constructingbinding-no-longer-needed-at-the-type-level
- Deprecate trailing slash


scavenger-collector/src/main/kotlin/com/navercorp/scavenger/config/JdbcConfig.kt
```diff
@@ -38,10 +38,9 @@ class JdbcConfig : AbstractJdbcConfiguration() {
         jdbcConverter: JdbcConverter
     ): SqlParameterSourceFactory {
         return EntityConvertibleSqlParameterSourceFactory(
-            ConvertibleParameterSourceFactory(DefaultJdbcParameterSourceConverter(), NoneFallbackParameterSource()),
+            ConvertibleParameterSourceFactory(DefaultJdbcParameterSourceConverter(Collections.emptyList()), NoneFallbackParameterSource()),
```
taeyeon-Kim, line 41: how about using `emptyList()` of kotlin??

scavenger-api/build.gradle.kts
```diff
@@ -3,9 +3,9 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import org.springframework.boot.gradle.tasks.bundling.BootJar
 
 plugins {
-    val kotlinVersion = "1.8.10"
-    val springBootVersion = "2.5.12"
-    val springDependencyManagementVersion = "1.0.11.RELEASE"
+    val kotlinVersion = "1.9.0"
```
taeyeon-Kim, line 6: Using the `1.9.24` version

taeyeon-Kim (commented):
Replace `io.github.microutils:kotlin-logging-jvm` with https://github.com/oshai/kotlin-logging

taeyeon-Kim (commented):
Delete `spring-cloud` if not needed

scavenger-api/build.gradle.kts
```diff
@@ -55,7 +55,7 @@ configure<DependencyManagementExtension> {
 }
 
 kotlin {
-    jvmToolchain(11)
+    jvmToolchain(17)
 }
 
 tasks.withType<KotlinCompile> {
```
taeyeon-Kim, line 61: maybe remove this option.

scavenger-collector/build.gradle.kts
```diff
@@ -37,20 +37,20 @@ dependencies {
...
-    implementation("com.linecorp.armeria:armeria-spring-boot2-starter")
-    implementation("com.linecorp.armeria:armeria-tomcat9")
+    implementation("com.linecorp.armeria:armeria-spring-boot3-starter")
+    implementation("com.linecorp.armeria:armeria-tomcat10")
     implementation("com.linecorp.armeria:armeria-grpc")
     implementation("io.github.microutils:kotlin-logging-jvm:2.1.23")
     implementation("net.ttddyy:datasource-proxy:1.7")
     implementation("com.github.vertical-blank:sql-formatter:1.0.1")
     testImplementation("org.springframework.boot:spring-boot-starter-test")
     testImplementation("org.mockito.kotlin:mockito-kotlin:3.2.0")
```
taeyeon-Kim, line 52: It is using mockito-core 5.7.0 on springboot 3.2.4.
Changing to https://mvnrepository.com/artifact/org.mockito.kotlin/mockito-kotlin/5.3.1

taeyeon-Kim (commented):
bump up `spring-boot-starter-data-jdbc` of `scavenger-entity` to `3.2.4`

scavenger-api/build.gradle.kts
sohyun-ku: fixed in https://github.com/naver/scavenger/pull/124/commits/576081e870da04e39edc820836090269dcec0a4a

scavenger-collector/build.gradle.kts
sohyun-ku: fixed in https://github.com/naver/scavenger/pull/124/commits/576081e870da04e39edc820836090269dcec0a4a

scavenger-collector/src/main/kotlin/com/navercorp/scavenger/config/JdbcConfig.kt
sohyun-ku: fixed in 1b356ae56e11c7a8305fb34d98c0898c8ee9ae5f

scavenger-collector/src/main/kotlin/com/navercorp/scavenger/repository/AgentStateDao.kt
```diff
@@ -68,4 +68,9 @@ class AgentStateDao(
                 .addValue("ids", ids)
         )
     }
+
+    fun toInstantMicro(instant: Instant): Instant {
+        val epochMicro = (instant.epochSecond * 1_000_000) + (instant.nano / 1000)
+        return Instant.ofEpochSecond(epochMicro / 1_000_000, (epochMicro % 1_000_000) * 1000)
+    }
```
sohyun-ku, line 75: After switching to jdk17, when using Instant.now(), I modified it to use micro because it tends to use different time precision depending on the OS.
That's why the test was previously crashed.

https://stackoverflow.com/questions/74781495/changes-in-instant-now-between-java-11-and-java-17-in-aws-ubuntu-standard-6-0

scavenger-api/build.gradle.kts
sohyun-ku: If we remove this option, it seems cannot detect jvm target like belows.
> w: Inconsistent JVM-target compatibility detected for tasks 'compileJava' (17) and 'compileKotlin' (1.8).

taeyeon-Kim: Have you tried deleting all build output, including model, entity, etc... and retrying?

scavenger-collector/src/main/kotlin/com/navercorp/scavenger/repository/AgentStateDao.kt
taeyeon-Kim: that's new to me, thanks 😄

scavenger-api/build.gradle.kts
```diff
@@ -50,12 +50,11 @@ dependencies {
 configure<DependencyManagementExtension> {
     imports {
         mavenBom(org.springframework.boot.gradle.plugin.SpringBootPlugin.BOM_COORDINATES)
```
taeyeon-Kim, line 52: Maybe this is also a configuration we need??

sohyun-ku: I think yes.
It allows us to leverage the Spring Boot BOM to simplify dependency management, ensure version consistency, and reduce the risk of version conflicts in our Spring Boot project.

> The SpringBootPlugin class provides a BOM_COORDINATES constant that can be used to import the bom without having to know its group ID, artifact ID, or version.
https://docs.spring.io/spring-boot/docs/3.2.4/gradle-plugin/reference/htmlsingle/#managing-dependencies.dependency-management-plugin.using-in-isolation

scavenger-api/build.gradle.kts
sohyun-ku: I deleted all the modules and retried, but they also crashed.
I think it would be better to mark it explicitly because there is a reason as below and there are cases where each modules use different versions of jdk.

> When there is no explicit information about the `jvmTarget` value in the build script, its default value is `null`, and the compiler translates it to the default value `1.8`
> https://kotlinlang.org/docs/gradle-configure-project.html#what-can-go-wrong-if-targets-are-incompatible

sohyun-ku: I think it would be better keep this option to run the service as before.
spring boot 3.x guide still includes `JSR305` option.

> https://github.com/spring-projects/spring-boot/blob/3.2.x/spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/kotlin.adoc#null-safety

taeyeon-Kim (approved):
👍 👍

scavenger-api/build.gradle.kts
```diff
@@ -50,17 +50,17 @@ dependencies {
...
 
 kotlin {
-    jvmToolchain(11)
+    jvmToolchain(17)
 }
 
 tasks.withType<KotlinCompile> {
     kotlinOptions {
         freeCompilerArgs = listOf("-Xjsr305=strict")
+        javaParameters = true
```
taeyeon-Kim, line 63: finally why was this option added?

sohyun-ku: for `spring-data-jdbc-plus`.
I refered this guide https://github.com/naver/spring-jdbc-plus/blob/29ef973d6fd1b66338683259dfd6423954110f00/guide-projects/plus-sql-java-kotlin-guide/build.gradle#L19

Issue #131, opened 2024-07-24
Bump up Java & Gradle


sohyun-ku:
Bump up Java & Gradle
- Java 17 -> 21
- Gradle 7.6.1 -> 8.9

sohyun-ku:
Job completed.
I'll be included in the next release.

PR #132, opened 2024-07-26
Bump up Java 21

sohyun-ku:
related to #131 
Bump up Java 21
- Build Results with Java 21 (on my local PC)

image

### DIFF
--- a/scavenger-old-agent-java/build.gradle.kts
+++ b/scavenger-old-agent-java/build.gradle.kts
@@ -36,6 +36,9 @@ dependencies {
 java {
     withJavadocJar()
     withSourcesJar()
+    toolchain {
+        languageVersion.set(JavaLanguageVersion.of(8))
+    }
 }
 
 tasks.compileJava {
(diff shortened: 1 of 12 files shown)


taeyeon-Kim (approved):
👍 👍

kojandy (approved):

scavenger-old-agent-java/build.gradle.kts
```diff
@@ -36,6 +36,9 @@ dependencies {
 java {
     withJavadocJar()
     withSourcesJar()
+    toolchain {
+        languageVersion.set(JavaLanguageVersion.of(8))
```
kojandy, line 40: What do you think about discontinuing Java 7 support from now on?

sohyun-ku: I think that's good.
If we keep supporting java 7, we'll have to give up more becuz we have to consider minimum java version compatibility.

PR #137, opened 2024-11-19
Implement class filtering using regular expressions

sohyun-ku:
related to #99 

An optional feature is provided to filter classes using regular expressions.
Filtering is based on the result of packages, excluding excludedPackages, and adding additionalPackages.


taeyeon-Kim (commented):
I think it's worth considering getting the `packages` setting as a regular expression.
What do you think of this suggestion?

sohyun-ku:
@taeyeon-Kim 

I agree with your suggestion.
What I was considering is we already have `additionalPackages`, `excludePackages` option.
so it seems necessary to prioritize them.

taeyeon-Kim:
Okay good, that's about the same level of setup as you said.
But how about `include~` rather than `exclude~`?? I think the negation has a downside that makes me think twice about it.

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/CodeBaseScanner.java
```diff
@@ -178,6 +184,11 @@ private boolean isAdditionalPackage(ClassNode clazz) {
         return additionalPackagePaths.stream().anyMatch(clazz.name::startsWith);
     }
 
+    private boolean isAdditionalByRegex(ClassNode clazz) {
+        String className = clazz.name.replaceAll("/", ".");
+        return additionalByRegex.stream().anyMatch(regex -> Pattern.matches(regex, className));
```
kojandy, line 189: In this approach, every time a class file is scanned, all regex strings in `additionalByRegex` and `excludeByRegex` will be recompiled to `Pattern`. Instead of holding them as `List<String>` in field, I think it would be better to pre-compile the regex patterns in constructor and store them as `List<Pattern>`.

> If a pattern is to be used multiple times, compiling it once and reusing it will be more efficient than invoking this method each time.
https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#matches-java.lang.String-java.lang.CharSequence-

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/ElementMatcherBuilder.java
```diff
@@ -55,11 +55,23 @@ public ElementMatcher<TypeDescription> buildClassMatcher() {
             .reduce(ElementMatcher.Junction::or)
             .orElse(none());
 
+        ElementMatcher.Junction<NamedElement> excludeByRegexMatcher = config.getExcludeByRegex().stream()
+            .map(ElementMatchers::nameMatches)
+            .reduce(ElementMatcher.Junction::or)
+            .orElse(none());
+
+        ElementMatcher.Junction<NamedElement> additionalByRegexMatcher = config.getExcludeByRegex().stream()
```
kojandy, line 63:
```suggestion
        ElementMatcher.Junction<NamedElement> additionalByRegexMatcher = config.getAdditionalByRegex().stream()
```

There is a typo here.

sohyun-ku: Oops.... fixed in 651d87debd035b4ab5499eb080271632d4cbfb87

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/CodeBaseScanner.java
sohyun-ku: As mentioned, `bytebuddy.matcher.StringMatcher.Mode.MATCHES` uses `Pattern.matches`. 
Therefore, switching to a List<Pattern> would improve performance.
Resolved in commit 0f630a6ded43036568165878dd1e0777a3512a2b.

taeyeon-Kim (approved):
👍 👍

kojandy (approved):

Issue #139, opened 2024-12-13
Replace MethodNode.parameters usage with Type.getArgumentTypes for improved stability


sohyun-ku:
The current code relies on the `MethodNode.parameters` field to determine method arguments. This field reads the MethodParameters information of the class file and sets the value, but only fills in if it contains a debug symbol (for example, compiled with the `-parameters` flag). In an environment where debug symbols are not available, this field may be null, which can lead to inconsistent behavior. (e.g. not a getter, but it is recognized as a getter due to missing parameters.)

To improve stability and ensure compatibility across different build environments, we should replace the usage of `MethodNode.parameters` with `Type.getArgumentTypes`. This approach directly analyzes the method descriptor and is independent of debug symbols.

- references
  - https://gitlab.ow2.org/asm/asm/-/blob/master/asm/src/main/java/org/objectweb/asm/MethodWriter.java#L450
  - https://dev.to/saladlam/get-class-method-parameter-name-from-class-file-at-runtime-2h2f

sohyun-ku:
This feature will be included in the next release.

PR #140, opened 2024-12-13
Replace MethodNode.parameters usage with Type.getArgumentTypes

sohyun-ku:
related to #139 

Replace usage of `MethodNode.parameters` with `Type.getArgumentTypes(methodNode.desc)`.
The code will work reliably in production environments where debug symbols may be stripped.


taeyeon-Kim (approved):
👍 👍

kojandy (approved):

PR #142, opened 2024-12-24
[WIP] Add optional MurmurHash support

kusalk:
The changes encompassed in this PR were originally included with those in #134. I extracted these change into this separate PR following a discussion in the former PR.

---

#### `MethodRegistry#getHash`
Scenario: `ConcurrentHashMap` cache disabled

| **Before PR (ns/op)**          | **After PR (ns/op)** |
| -------------------------------- | ---------------------- |
| 581.834 ± 46.560 | 139.283 ± 6.211       |

#### `InvocationTracker#onInvocation`
Scenario: Hash cache disabled, reset buffer state
The term 'reset buffer' refers to the state immediately after invocation data is published and the buffer is cleared, as opposed to 'not in buffer' which would refer to the state of the invocations buffer immediately upon agent startup.

| **Before PR (ns/op)**          | **After PR (ns/op)** |
| -------------------------------- | ---------------------- |
| TBA | 316.088 ± 36.943       |


scavenger-model/src/main/java/com/navercorp/scavenger/util/HashGenerator.java
```diff
@@ -1,37 +1,76 @@
...
+            return hashFunction.apply(signature);
         }
     }
 
-    public static class Md5 {
+    public static class DefaultHash {
+        private static final String SELECTED_ALGORITHM_PROP = "hash.algorithm";
+        private static final HashAlgorithm DEFAULT_ALGORITHM = HashAlgorithm.MD5;
+        private static final HashAlgorithm SELECTED_ALGORITHM =
+                HashAlgorithm.valueOf(System.getProperty(SELECTED_ALGORITHM_PROP, DEFAULT_ALGORITHM.name()).toUpperCase());
```
kusalk, line 34: We read the chosen hashing algorithm from the system property here

scavenger-model/build.gradle.kts
```diff
@@ -17,6 +22,7 @@ dependencies {
     implementation("io.grpc:grpc-kotlin-stub:${property("grpcKotlinVersion")}")
     implementation("io.grpc:grpc-protobuf:${property("grpcVersion")}")
     implementation("javax.annotation:javax.annotation-api:1.3.2")
+    implementation("commons-codec:commons-codec:1.17.1")
```
kusalk, line 25: We use the MurmurHash implementation from the `commons-codec` library dependency

scavenger-model/src/main/java/com/navercorp/scavenger/util/HashGenerator.java
```diff
@@ -1,37 +1,80 @@
...
+            md.update(signature.getBytes(StandardCharsets.UTF_8));
+            return String.format("%064x", new BigInteger(1, md.digest()));
+        }
+    }
+
+    private static class Md5 {
+        private static String from(String signature) {
+            MessageDigest md = callWithCheckedExceptionWrapping(() -> MessageDigest.getInstance("MD5"));
+            md.update(signature.getBytes(StandardCharsets.UTF_8));
+            // Ideally should be "%032x" to avoid confusion from leading zeros being dropped, do not change now, avoid compat issues
```
kusalk, line 63: I noticed this peculiarity whilst testing - the returned String won't always be 32 characters. This shouldn't cause issues but noted it in a comment. We obviously can't change it now anyway as it would cause backwards compatibility issues.

scavenger-model/build.gradle.kts
```diff
@@ -17,12 +22,19 @@ dependencies {
     implementation("io.grpc:grpc-kotlin-stub:${property("grpcKotlinVersion")}")
     implementation("io.grpc:grpc-protobuf:${property("grpcVersion")}")
     implementation("javax.annotation:javax.annotation-api:1.3.2")
+    implementation("commons-codec:commons-codec:1.18.0")
```
kusalk, line 25: We use the MurmurHash implementation from the `commons-codec` library dependency - it will increase the size of the agent JAR slightly (~300 kB)

PR #148, opened 2025-04-11
Update SampleApp

sohyun-ku:
I added this because the integrationTest sometimes fails intermittently due to the sampleApps shutting down too quickly, even before they receive a response from the collector.


taeyeon-Kim (approved):

taeyeon-Kim (commented):
This can slow down integration tests. Is this acceptable?

sohyun-ku:
@taeyeon-Kim 

As you mentioned, these changes may slow down the integration tests.
However, I believe consistency in test results is more important.
Also, since integration tests only run on CI servers, there’s no impact on the actual production environment, so I think this trade-off is acceptable.

cc. @kojandy

kojandy (approved):

PR #150, opened 2025-04-21
Implement method call stack tracing

### LINKED ISSUE #109: implement method call stack
For deeper analysis of dead code (for example, to recognize that a method you've thought was dead is being called and you want to know what it is), we also support the feature that enables you to see the method call stack.

sohyun-ku:
- resolved: #109 
- **Call Stack Tracing with ConcurrentHashMap + Deque**
  - To manage call stacks per thread, we used the structure `ConcurrentHashMap<Long, ArrayDeque<String>>.`
  - Each thread maintains its own call stack (`Deque`), and method entries and exits are tracked via `push` and `pop` operations respectively.
  - The `Deque` is used as a stack to accurately capture the order and depth of method calls.
- **Caller-Callee Based Lightweight Call Structure**
  - Instead of storing the entire method call stack, each call holds a reference to its direct caller only.
  - This design retains the hierarchical relationship between method calls while significantly reducing memory overhead by avoiding redundant stack information.
  - It allows for a clear understanding of “who called whom” at the direct (1-depth) level,
and can be effectively used for incremental exploration or future enhancements such as call relationship visualization.
- **Benchmark Results (`ns/op`)**
  - The benchmarking tests were conducted on a local machine with an Apple M1 Pro, 32GB RAM. Please note that results may vary depending on the environment.
  - Since enabling `CallStackTraceMode` introduces some overhead, the feature is implemented as optional, allowing it to be turned on only when detailed call tracing is needed.

  Java version | CallStackTraceMode False | CallStackTraceMode True
  -- | -- | --
  Java 8 | 14.056 ± 0.235 | 34.004 ± 0.403
  Java 11 | 12.933 ± 0.247 | 27.244 ± 0.344
  Java 17 | 12.831 ± 0.340 | 26.625 ± 0.248
  Java21 | 12.502 ± 0.104 | 26.408 ± 3.061

- **Snapshot: Caller-Callee Relationship Example**

스크린샷 2025-04-21 오후 4 53 52


scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/CallStackRegistry.java
```diff
@@ -0,0 +1,48 @@
+package com.navercorp.scavenger.javaagent.collecting;
+
+import com.navercorp.scavenger.javaagent.model.Config;
+import com.navercorp.scavenger.model.CallStackDataPublication;
+
+import lombok.extern.java.Log;
+
+import java.util.*;
```
taeyeon-Kim: How about importing them one by one instead of with wildcard?

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/CallStackRegistry.java
```diff
@@ -0,0 +1,48 @@
...
+        callers.add(caller);
+    }
+
+    public CallStackDataPublication getPublication(Config config, String codeBaseFingerprint) {
+        Set<CallStackDataPublication.CallStackDataEntry> dataEntries = new HashSet<>();
+        for (Map.Entry<String, Set<String>> entry : callStacks.entrySet()) {
+            String callee = entry.getKey();
+            Set<String> callers = entry.getValue();
+            if (!callers.isEmpty()) {
+                CallStackDataPublication.CallStackDataEntry CallStackDataEntry = CallStackDataPublication.CallStackDataEntry.newBuilder()
```
taeyeon-Kim:
```suggestion
                CallStackDataPublication.CallStackDataEntry callStackDataEntry = CallStackDataPublication.CallStackDataEntry.newBuilder()
```

Fix typo 😄

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/CallStackRegistry.java
```diff
@@ -0,0 +1,48 @@
...
+    private long recordingIntervalStartedAtMillis = System.currentTimeMillis();
+
+    public void register(String caller, String callee) {
+        Set<String> callers = callStacks.computeIfAbsent(callee, k -> new HashSet<>());
+        callers.add(caller);
+    }
+
+    public CallStackDataPublication getPublication(Config config, String codeBaseFingerprint) {
+        Set<CallStackDataPublication.CallStackDataEntry> dataEntries = new HashSet<>();
+        for (Map.Entry<String, Set<String>> entry : callStacks.entrySet()) {
```
taeyeon-Kim:
```java
callStacks.forEach((callee, callers) -> {
```
forEach enables pre-declaration.

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/CallStackTracker.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public CallStackTracker(CallStackRegistry callStackRegistry, MethodRegistry methodRegistry, boolean isDebugMode) {
+        this.callStackRegistry = callStackRegistry;
+        this.methodRegistry = methodRegistry;
+        this.isDebugMode = isDebugMode;
+        INSTANCE = this;
+    }
+
+    public void installAdvice(Instrumentation inst, Config config) {
+        ElementMatcherBuilder matcherBuilder = new ElementMatcherBuilder(config.copyConfigWithExcludeConstructorsTrue());
```
taeyeon-Kim, line 54: Is there a reason to copy `excludeConstructors` to true?

scavenger-agent-java/src/test/java/com/navercorp/scavenger/javaagent/scheduling/SchedulerTest.java
```diff
@@ -81,7 +88,9 @@ public class SchedulerTest {
     public void setUp() throws Exception {
         lenient().when(codeBaseScannerMock.scan())
             .thenReturn(new CodeBase(List.of(Method.createTestMethod()), "fingerprint"));
-        sut = new Scheduler(invocationRegistry, new Config(new Properties()), publisher, codeBaseScannerMock);
+        Properties properties = new Properties();
+        properties.setProperty("callStackTraceMode", "true");
```
taeyeon-Kim, line 96: To make it more clear, it also needs to be tested when `callStackTraceMode` is `false`

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/scheduling/Scheduler.java
```diff
@@ -222,4 +243,28 @@ public void publishInvocationDataIfNeeded() {
             }
         }
     }
+
+    public void publishCallStackDateIfNeeded() {
+        if (config.isCallStackTraceMode() && callStackDataPublisherState.isDueTime() && dynamicConfig != null && isCodeBasePublished) {
```
taeyeon-Kim: this method is better to just focus on publish. How about checking the CallStackTraceMode on the caller?

scavenger-api/src/main/kotlin/com/navercorp/scavenger/repository/sql/CallStackSql.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    fun selectCallerSignatures(
+        customerId: Long,
+        applicationIds: List<Long>,
+        environmentIds: List<Long>,
+        signature: String,
+        invokedAtMillis: Long?,
+    ): String =
+        """
+        SELECT
```
taeyeon-Kim, line 9: Can you check query performance?
```
EXPLAIN FORMAT=VITESS {query}
```

taeyeon-Kim (commented):
Sorry for the late review. I only checked the agent and api modules first.
- Is it possible to see the call stack tree maybe?
- Also, could you please change the vitess configuration file.

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/CallStackTracker.java
sohyun-ku: ByteBuddy does not support exception hooking for constructors, which can lead to an `IllegalStateException`, so the option was not used.

- refer: `https://github.com/raphw/byte-buddy/issues/1420`

scavenger-collector/src/main/kotlin/com/navercorp/scavenger/controller/GrpcAgentController.kt
```diff
@@ -76,6 +91,19 @@ class GrpcAgentController(
         validate(request.commonData)
     }
 
+    private fun validate(request: CallStackDataPublication) {
+        if (!request.hasCommonData()) {
+            throw IllegalArgumentException("CommonPublicationData is a mandatory field")
+        }
+        for (entry in request.entryList) {
+            if (entry.callee.isEmpty() || entry.callersList.isEmpty()) {
```
taeyeon-Kim: How about different exception messages for separate conditions?

scavenger-api/src/main/kotlin/com/navercorp/scavenger/repository/sql/CallStackSql.kt
```diff
@@ -0,0 +1,28 @@
+package com.navercorp.scavenger.repository.sql
+
+import com.navercorp.spring.data.jdbc.plus.sql.support.SqlGeneratorSupport
+
+class CallStackSql : SqlGeneratorSupport() {
+
+    fun selectCallerSignatures(
```
taeyeon-Kim: parameter is probably not needed.

scavenger-collector/src/main/kotlin/com/navercorp/scavenger/repository/MethodDao.kt
```diff
@@ -59,7 +59,7 @@ class MethodDao(
             return 0
         }
         return update(
-            sql.deleteAllMethodsAndInvocations(),
+            sql.deleteAllMethods(),
```
taeyeon-Kim, line 62: 👍

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/CallStackTracker.java
taeyeon-Kim: Isn't this the same problem with invocation?

sohyun-ku: This is specifically related to the `onThrowable = Throwable.class`.
Therefore, this isn’t an issue for invocations, as they don’t involve exception handling.

taeyeon-Kim: understand., thanks for the explanation.

sohyun-ku:
@taeyeon-Kim 

> Is it possible to see the call stack tree maybe?

In the current PR, the implementation is based on a one-way `caller-callee` relationship. However, I expect that in future feature expansions, it will be possible to construct the full call tree by recursively tracing the call path in reverse from a specific `callee`.
I considered various ways to model the call structure, but ultimately chose this simple caller-callee approach for its simplicity and ease of maintenance.
If you see any limitations in the current structure or believe another approach might be more appropriate, I’d appreciate your feedback.

taeyeon-Kim:
@sohyun-ku This is a great idea, and for an early version, this is great!!
Can I ask you to check one thing, if it is possible to extend this, for example, to make it possible to draw a tree from Caller to Root Caller.

sohyun-ku:
@taeyeon-Kim 

I think it is technically possible to build a tree from a specific callee up to the root
(e.g., by traversing the reversed graph recursively to explore all possible call paths).

However, the stored call data represents multiple call records across the application rather than a single call flow.
Therefore, the constructed tree resembles a combination of all possible call relationships rather than an actual call stack.

Additionally, since the call graph may contain cycles and paths can get deep,
it’s important to apply constraints such as a visited set to avoid duplicates and depth limits to prevent excessive recursive traversal.

In summary, while extending the tree up to the root is feasible, please note that this does not exactly reproduce a single actual call flow that triggered the callee.

Due to the necessity of minimizing the performance impact on the agent,
these constraints inevitably arose to reduce performance and memory overhead during tree construction.

taeyeon-Kim (approved):
🔥 👍 👍 👍 👍

taeyeon-Kim:
> @taeyeon-Kim
> 
> I think it is technically possible to build a tree from a specific callee up to the root (e.g., by traversing the reversed graph recursively to explore all possible call paths).
> 
> However, the stored call data represents multiple call records across the application rather than a single call flow. Therefore, the constructed tree resembles a combination of all possible call relationships rather than an actual call stack.
> 
> Additionally, since the call graph may contain cycles and paths can get deep, it’s important to apply constraints such as a visited set to avoid duplicates and depth limits to prevent excessive recursive traversal.
> 
> In summary, while extending the tree up to the root is feasible, please note that this does not exactly reproduce a single actual call flow that triggered the callee.
> 
> Due to the necessity of minimizing the performance impact on the agent, these constraints inevitably arose to reduce performance and memory overhead during tree construction.

@sohyun-ku yes I understand, thanks for the clarification. 😄

scavenger-api/src/main/kotlin/com/navercorp/scavenger/repository/sql/CallStackSql.kt
sohyun-ku: Sorry for the late reply :(
Here is the Vitess query plan
```
+----------+-------------------+----------+-------------+------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| operator | variant           | keyspace | destination | tabletType | query                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 |
+----------+-------------------+----------+-------------+------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Route    | SelectEqualUnique | main     |             | UNKNOWN    | select caller_methods.signature as callerSignature from call_stacks join methods as callee_methods on call_stacks.customerId = callee_methods.customerId and call_stacks.signatureHash = callee_methods.signatureHash join methods as caller_methods on call_stacks.customerId = caller_methods.customerId and call_stacks.callerSignatureHash = caller_methods.signatureHash where callee_methods.signature = '${PATH}' and call_stacks.customerId = 11000002003 and call_stacks.applicationId in (11003142347) and call_stacks.environmentId in (11003142346) and (1751941260000 is null or call_stacks.invokedAtMillis >= 1751941260000) |
+----------+-------------------+----------+-------------+------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/CallStackRegistry.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+@Log
+public class CallStackRegistry {
+    private final Map<String, Set<String>> callStacks = new ConcurrentHashMap<>();
+    private long recordingIntervalStartedAtMillis = System.currentTimeMillis();
+
+    public void register(String caller, String callee) {
+        Set<String> callers = callStacks.computeIfAbsent(callee, k -> new HashSet<>());
+        callers.add(caller);
```
kojandy: `HashSet` is not thread-safe, so there could be some issues if `register()` is called concurrently with `getPublication()`. Maybe consider using `ConcurrentHashMap.newKeySet()` or another thread-safe data structure.

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/CallStackTracker.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Advice.OnMethodExit(onThrowable = Throwable.class)
+    public static void onExit() {
+        updateCallStackOnExit();
+    }
+
+    public static void updateCallStackOnExit() {
+        ArrayDeque<String> currentThreadCallStack = CALL_STACKS.get(Thread.currentThread().getId());
+        String signature = currentThreadCallStack.pollLast();
```
kojandy: In contrast to the `HashSet` situtation above, `ArrayDeque` doesn't need to be thread-safe here since access to each `ArrayDeque` is seperated by `threadId`. However, I think there should be some sort of check whether the `ArrayDeque` is empty after `pollLast()`. Otherwise, empty `ArrayDeque` will remain in memory even after thread termination. I think we should remove empty `ArrayDeque` explicitly or use `ThreadLocal` instead.

kojandy (commented):
Sorry for the late review... I wanted to test more thoroughly by running it locally, but wasn't able to do as much. Really appreciate all the hard work you put into this! 🔥🔥

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/CallStackTracker.java
sohyun-ku: As you mentioned, accumulating empty objects could impact memory.
However, using ThreadLocal or removing them after every call might affect performance,
so I implemented the cleanup using a scheduler instead(6c307f5)

sohyun-ku:
Here are the final benchmark performance test results for the current implementation 
(measured on an Apple M3 Pro with 32 GB memory), after adding a few more commits:

| | CallStackTraceMode False | CallStackTraceMode True|
-- | -- | --
Java 8 | 6.529 ± (99.9%) 0.457 | 17.183 ± (99.9%) 0.349
Java 11 | 5.636 ± (99.9%) 0.425 | 16.381 ± (99.9%) 2.242
Java 17 | 5.825 ± (99.9%) 0.843 | 16.634 ± (99.9%) 2.209
Java 21 | 5.309 ± (99.9%) 0.137 | 17.348 ± (99.9%) 3.335

</div></div>

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/CallStackRegistry.java
sohyun-ku: fixed in 364dae1

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/CallStackTracker.java
```diff
@@ -0,0 +1,139 @@
...
+        AgentBuilder transform = new AgentBuilder.Default(new ByteBuddy().with(TypeValidation.DISABLED))
+            .type(matcherBuilder.buildClassMatcher())
+            .transform((builder, typeDescription, classLoader, module, protectionDomain) ->
+                builder.visit(advice.on(matcherBuilder.buildMethodMatcher(typeDescription)))
+            );
+        if (isDebugMode) {
+            transform = transform.with(new AgentBuilder.Listener.Adapter() {
+                @Override
+                public void onTransformation(
+                    @NotNull TypeDescription typeDescription,
```
taeyeon-Kim: It's using `org.jetbrains.annotations.NotNull`. Is this what you expect?

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/CallStackTracker.java
```diff
@@ -0,0 +1,139 @@
...
+    @Advice.OnMethodEnter
+    public static void onEnter(@Advice.Origin String signature) {
+        saveCallTraceOnEnter(signature);
+    }
+
+    public static void saveCallTraceOnEnter(String signature) {
+        ArrayDeque<String> currentThreadCallStack = CALL_STACKS.computeIfAbsent(Thread.currentThread().getId(), k -> new ArrayDeque<>());
+        String callee = INSTANCE.methodRegistry.getHash(signature);
+
+        if (!currentThreadCallStack.isEmpty()) {
```
taeyeon-Kim, line 87: Is it possible that `currentThreadCallStack` will be null caused by the `startCallStackCleanup`?

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/CallStackTracker.java
sohyun-ku: It seems this annotation was automatically added by the IDE.
I’ve removed it to keep consistency with other parts of the code.

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/CallStackTracker.java
sohyun-ku: In the ideal flow, the value should always be present, but due to concurrency issues, problems might occur. Just in case, I’ve added a null check as a precaution (e.g., when the call depth is 1).

taeyeon-Kim (approved):
👍 👍

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/collecting/CallStackTracker.java
kojandy: After creating an empty `ArrayDeque` via `computeIfAbsent()`  and before adding the callee with `addLast()`, the cleanup thread might remove the empty `ArrayDeque`.

sohyun-ku: @kojandy 
Such issues, as you mentioned, may occur, but I felt prioritizing performance made sense, so I chose this approach.
Maybe related to this discussion https://github.com/naver/scavenger/pull/150#discussion_r2268806634 .
But if you think differently, feel free to let me know anytime!

kojandy: Oh, understood. seems reasonable!

kojandy (approved):
🙏🙏

PR #159, opened 2025-09-04
Display Full Call Stack on snapshot

sohyun-ku:
- related : #109 
- Changes
  1. Prevent Errors on Empty Call Stack
  2. Refactored FE to Expose the Full Call Stack
      - Backend now returns the entire Call Stack Map.
      - The map is stored in the Vue store on the frontend.
      - When a signature is clicked, the frontend generates the Call Stack Trace from the stored data.
      - This reduces the need for the server to generate traces on demand, lowering server load.

image


kojandy (approved):
👍

taeyeon-Kim (approved):
👍 👍

PR #161, opened 2025-09-15
Modify VSchema rollback previous changes and add new version

sohyun-ku:
- related : #150
- Changes
  - Added a new versioned VSchema file for continued schema migration
  - Ensures traceability and provides a safer rollout path for future changes

### DIFF
--- a/scavenger-schema/vitess/vschema_shared-1.1.4.json
+++ b/scavenger-schema/vitess/vschema_shared-1.1.4.json
@@ -153,6 +153,18 @@
                     "name": "hash"
                 }
             ]
+        },
+        "call_stacks": {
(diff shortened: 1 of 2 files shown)


taeyeon-Kim (approved):
Since configuration updates in vitess can overwrite configurations, it seems Vschema doesn't need versioning. how about that??

sohyun-ku:
@taeyeon-Kim 

As mentioned, VSchema is usually overwritten, so versioning isn’t strictly required.  
This change ensures safe application for users with versions earlier than 1.1.4  
and adds a new file for history tracking.

kojandy (approved):

PR #167, opened 2026-03-10
MCP

tngusdl3719:
This PR adds an MCP (Model Context Protocol) server to Scavenger API, enabling AI assistants (Claude, Cursor, etc.) to query runtime code usage data directly via natural language.
                                                                                                                                                                                                                                              
  - Expose Scavenger's analysis capabilities as MCP tools over Streamable HTTP (`/mcp`)                                                                                                                                                       
  - Support 17 tools covering workspaces, snapshots, dead code analysis, and cleanup recommendations                                                                                                                                          
  - Auto-detect workspace from git repository URL via `resolveWorkspace`                                                                                                                                                                      
  - Identify methods that stopped being called after the last deployment via `getMethodsNotInvokedSinceLastDeploy`
  - Add CORS filter for MCP endpoint to support browser-based clients (MCP Inspector, etc.)
  - Add `doc/mcp-guide.md` with setup instructions and example prompts

  ## New MCP Tools

  | Tool | Description |
  |------|-------------|
  | `listCustomers` | List all registered workspaces |
  | `listApplications` / `listEnvironments` / `listSnapshots` | Browse workspace resources |
  | `getSnapshotUsageSummary` | Used/unused method counts per snapshot |
  | `getSnapshotTree` | Browse hierarchical code tree |
  | `searchMethods` / `isMethodUsed` | Search and check method usage |
  | `getDeadCode` | Get all unused methods |
  | `getClassUsageSummary` / `getPackageUsageSummary` | Usage breakdown by class or package |
  | `analyzeCleanupCandidates` | Dead code ranked by removal confidence |
  | `getMethodsNotInvokedSinceLastDeploy` | Newly dead code since last deploy |
  | `getUsageOverview` | High-level workspace usage stats |
  | `createSnapshot` / `refreshSnapshot` | Manage snapshots |
  | `resolveWorkspace` | Find workspace by git URL |

  ## How to Connect

  **Claude Desktop** — add to `claude_desktop_config.json`:
  ```json
  {
    "mcpServers": {
      "scavenger": {
        "type": "http",
        "url": "http://localhost:8081/scavenger/mcp"
      }
    }
  }

  MCP Inspector (testing):
  npx @modelcontextprotocol/inspector
  Set transport to Streamable HTTP and enter the URL above.

  Test Plan

  - Start Scavenger API and verify /mcp endpoint responds
  - Connect via MCP Inspector — confirm all 17 tools are listed
  - Run listCustomers → listSnapshots → getDeadCode flow
  - Connect via Claude Desktop and run a natural language dead code query
  - Verify CORS headers are present on /mcp responses


taeyeon-Kim (commented):
@tngusdl3719 First, thank you for the great PR.
Before that, could the test implementation be done for the features you added??

tngusdl3719:
> @tngusdl3719 First, thank you for the great PR. Before that, could the test implementation be done for the features you added??

Sure, I’ll add the test implementation for the features included in this PR. I’ll update the PR once the tests are ready.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/mcp/ScavengerMcpService.kt
```diff
@@ -0,0 +1,360 @@
...
+    private val snapshotService: SnapshotService,
+    private val snapshotNodeService: SnapshotNodeService,
+    private val snapshotNodeDao: SnapshotNodeDao,
+    private val summaryService: SummaryService,
+    private val methodRepository: MethodRepository,
+    private val githubMappingRepository: GithubMappingRepository,
+    private val customerRepository: CustomerRepository,
+    private val jvmRepository: JvmRepository,
+) {
+    fun listCustomers(): List<Map<String, Any>> {
```
taeyeon-Kim: `CustomerService.getCustomers()` already returns `List<CustomerDto>` with the exact same fields (`id`, `name`). The same pattern repeats throughout this file:

| Method | Existing service return type | What this PR does |
|---|---|---|
| `listCustomers()` | `List<CustomerDto>` | Re-maps to `mapOf("id" to it.id, "name" to it.name)` |
| `listApplications()` | `List<ApplicationDetailDto>` | Re-maps to `mapOf("id", "name", "jvmCount", ...)` |
| `listEnvironments()` | `List<EnvironmentDetailDto>` | Re-maps to `mapOf("id", "name", "jvmCount", ...)` |
| `listSnapshots()` | `List<SnapshotDto>` | Re-maps to `mapOf("id", "name", "createdAt", ...)` |
| `createSnapshot()` | `SnapshotDto` | Re-maps to `mapOf("id", "name", "createdAt", ...)` |
| `getUsageOverview()` | `SummaryDto` | Re-maps to `mapOf("totalMethodsTracked", ...)` |

Just return the existing DTOs directly — `ScavengerMcpTools.toJson()` will serialize them via Jackson.

For MCP-specific responses (`analyzeCleanupCandidates`, `getClassUsageSummary`, etc.), please define new DTOs in the `dto/` package following the `data class` + `companion object { fun from() }` pattern used throughout this project.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/mcp/ScavengerMcpService.kt
```diff
@@ -0,0 +1,360 @@
...
+            "customerName" to customerName,
+            "totalMethodsTracked" to summary.methodCount,
+            "snapshotCount" to snapshots.size,
+            "snapshotLimit" to summary.snapshotLimit,
+            "applications" to apps.map { mapOf("name" to it.name, "invocationCount" to it.invocationCount) },
+            "environments" to envs.map { mapOf("name" to it.name) }
+        )
+    }
+
+    private fun SnapshotNodeEntity.toMap(): Map<String, Any?> = mapOf(
```
taeyeon-Kim: This project uses `Dto.from(entity)` factory methods for entity-to-response conversion (see `SnapshotDto.from()`, `CustomerDto.from()`, etc.). Please follow the same pattern instead of `Entity.toMap()` extension functions.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/mcp/ScavengerMcpService.kt
```diff
@@ -0,0 +1,360 @@
...
+        limit: Int
+    ): Map<String, Any?> {
+        val customerId = customerResolver.resolveCustomerId(customerName)
+        val cappedLimit = limit.coerceIn(1, 500)
+
+        val deployTime = if (sinceMillis != null) {
+            sinceMillis
+        } else {
+            val latestPublishedAt = jvmRepository.findLatestPublishedAt(customerId)
+                ?: return mapOf(
```
taeyeon-Kim: Mixing error and success in the same `Map<String, Any?>` return type is fragile — callers must check `result["error"]` to branch. Consider throwing an exception (consistent with how `McpCustomerResolver` handles missing customers) or using a sealed class.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/mcp/McpCustomerResolver.kt
```diff
@@ -0,0 +1,18 @@
+package com.navercorp.scavenger.mcp
+
+import com.navercorp.scavenger.repository.CustomerRepository
+import org.springframework.stereotype.Component
+
+@Component
+class McpCustomerResolver(
```
taeyeon-Kim: `CustomerService.getCustomerByName(groupId, customerName)` already does the same lookup:
```kotlin
fun getCustomerByName(groupId: String, customerName: String): CustomerDto {
    return customerRepository.findByNameAndGroupId(customerName, groupId)
        .orElseThrow { EmptyResultDataAccessException(1) }
        .let { CustomerDto.from(it) }
}
```
Consider using `CustomerService` directly instead of introducing a new resolver class.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/mcp/McpCorsFilter.kt
```diff
@@ -0,0 +1,42 @@
...
+import jakarta.servlet.FilterChain
+import jakarta.servlet.ServletRequest
+import jakarta.servlet.ServletResponse
+import jakarta.servlet.http.HttpServletRequest
+import jakarta.servlet.http.HttpServletResponse
+import org.springframework.boot.web.servlet.FilterRegistrationBean
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+
+@Configuration
```
taeyeon-Kim: The project already has `WebConfig.kt` implementing `WebMvcConfigurer`. CORS can be added there instead of a raw servlet filter:
```kotlin
override fun addCorsMappings(registry: CorsRegistry) {
    registry.addMapping("/mcp/**")
        .allowedOrigins("*") // should be configurable
        .allowedMethods("GET", "POST", "DELETE", "OPTIONS")
        .exposedHeaders("Mcp-Session-Id")
}
```
Also, hardcoding `Access-Control-Allow-Origin: *` is a security concern — consider making it configurable via properties.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/repository/JvmRepository.kt
```diff
@@ -12,6 +12,9 @@ interface JvmRepository : DelegatableJdbcRepository<JvmEntity, String> {
     fun countByCustomerIdAndApplicationId(customerId: Long, applicationId: Long): Long
     fun countByCustomerIdAndEnvironmentId(customerId: Long, environmentId: Long): Long
 
+    @Query("SELECT MAX(j.publishedAt) FROM jvms j WHERE j.customerId = :customerId")
+    fun findLatestPublishedAt(@Param("customerId") customerId: Long): java.time.Instant?
```
taeyeon-Kim:
```suggestion
    fun findLatestPublishedAt(@Param("customerId") customerId: Long): Instant?
```
Please use an import instead of FQCN (`java.time.Instant`). This is a project coding standard.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/repository/SnapshotNodeDao.kt
```diff
@@ -7,6 +7,13 @@ import com.navercorp.spring.data.jdbc.plus.sql.provider.EntityJdbcProvider
 import com.navercorp.spring.data.jdbc.plus.sql.support.JdbcDaoSupport
 import org.springframework.stereotype.Repository
 
+data class SnapshotMethodUsageSummary(
```
taeyeon-Kim: Data classes used as query results should not be defined inside a DAO file. Move to the `dto/` package or as a dedicated DB row class — see `SnapshotExportDbRow` for the existing pattern.

scavenger-api/build.gradle.kts
```diff
@@ -18,11 +18,14 @@ plugins {
 
 repositories {
     mavenCentral()
+    maven { url = uri("https://repo.spring.io/milestone") }
```
taeyeon-Kim: `spring-ai-starter-mcp-server-webmvc:1.1.2` is a GA release available on Maven Central. This milestone repository is unnecessary — please remove it.

scavenger-api/src/test/kotlin/com/navercorp/scavenger/mcp/ScavengerMcpToolsTest.kt
```diff
@@ -0,0 +1,119 @@
...
+    @Test
+    fun listCustomers() {
+        val result = sut.listCustomers()
+        assertThat(result).contains("demo")
+    }
+
+    @Test
+    fun listApplications() {
+        val result = sut.listApplications(customerName)
+        assertThat(result).contains("name")
```
taeyeon-Kim: These assertions only verify that a JSON string contains a keyword:
```kotlin
assertThat(result).contains("name")  // passes even if "name" appears in a value
```
This doesn't validate response structure at all. Consider deserializing the JSON and asserting on typed fields. Or, since `ScavengerMcpServiceTest` already has more thorough assertions, consider whether this test class adds meaningful coverage beyond that.

taeyeon-Kim (changes requested):
Thanks for the contribution! The MCP server idea is valuable and the tool coverage is comprehensive.

However, there are several convention mismatches with the existing codebase that need to be addressed before merging. Most issues come from not reusing existing services, DTOs, and patterns already established in the API module.

### Must fix
1. **Use typed DTOs instead of `Map<String, Any?>`** — Every existing service returns typed `data class` DTOs (see `dto/` package). Returning raw Maps loses type safety, breaks IDE support, and diverges from the project convention.
2. **Reuse existing services and DTOs** — `CustomerDto`, `ApplicationDetailDto`, `EnvironmentDetailDto`, `SnapshotDto`, `SummaryDto` already exist. Most of `ScavengerMcpService` is re-mapping data that these DTOs already represent.
3. **FQCN usage** — Use imports instead of fully qualified class names (project standard).

### Should fix
4. **`SnapshotMethodUsageSummary` location** — Data classes should not be defined inside DAO files. Move to `dto/` package.
5. **CORS filter** — Use Spring's `WebMvcConfigurer.addCorsMappings()` (already used in `WebConfig.kt`) instead of a raw servlet filter. `Access-Control-Allow-Origin: *` should be configurable.
6. **`McpCustomerResolver` duplicates `CustomerService`** — `CustomerService.getCustomerByName()` already does the same lookup.
7. **Weak test assertions in `ScavengerMcpToolsTest`** — Checking `assertThat(jsonString).contains("name")` doesn't validate response structure.

### Nit
8. **Unnecessary milestone repository** — `spring-ai 1.1.2` is a GA release on Maven Central. Remove `repo.spring.io/milestone`.
9. **`ScavengerMcpProperties` mixed with config class** — Separate into its own file per project convention.

taeyeon-Kim:
**Additional review based on the intended design direction for this feature.**

We have design decisions for the MCP server that differ from this PR's approach. Here are the key points:

### 1. Authentication: apiKey-based customer scoping (Critical)

Scavenger agents already authenticate via `apiKey` (= `licenseKey` in the DB). Each apiKey is bound to a specific customer (workspace). The MCP server should use the same mechanism:

- Client passes apiKey (e.g. via `X-Scavenger-Api-Key` header)
- Server calls `LicenseService.check(apiKey)` → resolves `customerId`
- All subsequent tool calls are scoped to that `customerId` only
- **`customerName` parameter should be removed from all tools** — customer is determined by the apiKey

Currently, the PR exposes all customers via `listCustomers()` with no authentication. Anyone with access to the `/mcp` endpoint can query any workspace's data.

### 2. Read-only tools only — remove mutation tools (Major)

`createSnapshot` and `refreshSnapshot` modify data. MCP tools should be **read-only** — snapshot management belongs in the web UI. Please remove these.

### 3. Too many tools — focus on core scenarios (Major)

The core MCP use cases are: (1) check if a method is used, (2) list dead code, (3) get usage summary, (4) check who calls a method, (5) list snapshots, (6) find newly dead code since deploy.

Tools that become unnecessary with apiKey auth: `listCustomers`, `resolveWorkspace`.
Tools that overlap with the above: `searchMethods`, `getClassUsageSummary`, `getPackageUsageSummary`, `getSnapshotTree`.
Tools that should not exist: `listApplications`, `listEnvironments` (not relevant to MCP use cases).

**Missing**: a `get_method_callers` tool using callstack data — this is the most valuable tool for safe deletion decisions, as it tells you whether removing a dead method will break any caller.

### 4. Automatic snapshotId resolution (Major)

Most tools currently require an explicit `snapshotId`, which forces the AI to call `listSnapshots` first on every interaction. When `snapshotId` is omitted, the server should automatically use the latest snapshot. This significantly reduces tool call overhead.

### 5. Standardized error responses (Minor)

Please define a consistent error format (e.g. `{ "error": "SNAPSHOT_NOT_FOUND", "message": "...", "hint": "..." }`) instead of mixing error state into `Map<String, Any?>` with `"error" to true`.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/mcp/ScavengerMcpService.kt
tngusdl3719: Thanks. I aligned the MCP layer with that direction in the latest update.

`listCustomers()`, `listApplications()`, `listEnvironments()`, `listSnapshots()`, and `createSnapshot()` now return the existing typed DTOs directly instead of remapping them into `Map<String, Any?>`.

For MCP-only responses such as cleanup analysis, class/package usage summaries, snapshot tree nodes, workspace resolution, and deploy-based method analysis, I added dedicated DTOs under dto/.

`getUsageOverview()` now reuses SummaryDto as part of the response and wraps the additional MCP-specific fields in a typed MCP DTO rather than flattening everything into a raw map.

One nuance is that not all DTOs in this codebase use `companion object { fun from() }`, especially for responses composed from multiple sources rather than a single entity. I followed the existing DTO style where reuse made sense, and kept the MCP-specific aggregation DTOs as typed data classes.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/mcp/ScavengerMcpService.kt
tngusdl3719: Addressed.
I moved the DTO mappings that were being created directly from SnapshotNodeEntity into `Dto.from(entity)` factory methods.

Added/updated:
- `McpSnapshotNodeDto.kt`
- `McpMethodUsageDto.kt`
- `McpMethodReferenceDto.kt`

On the service side, I replaced the extension function calls with Mcp...`Dto.from(...)`, and removed the remaining entity-to-DTO extension mappers from `ScavengerMcpService.kt`.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/mcp/ScavengerMcpService.kt
tngusdl3719: Thanks. I updated `getMethodsNotInvokedSinceLastDeploy()` to stop mixing success and error states in the same response shape.

- The method now throws when deploy metadata is unavailable instead of returning a partially-populated success DTO with error/message fields.
- `McpMethodsNotInvokedSinceLastDeployResultDto` is now a success-only DTO.
- I also updated the corresponding test to assert the exception path.

This keeps the MCP response contract consistent with the rest of the service layer and avoids requiring callers to branch on an error flag inside an otherwise successful response object.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/mcp/McpCustomerResolver.kt
tngusdl3719: I removed McpCustomerResolver and now use CustomerService.getCustomerByName(...) directly, while keeping the MCP-specific error message handling in ScavengerMcpService.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/mcp/McpCorsFilter.kt
tngusdl3719: Addressed. I removed the raw servlet filter and moved MCP CORS configuration into `WebConfig.addCorsMappings()`, and the allowed origins are now configurable via `scavenger.mcp.allowed-origins`. The current default remains `"*"` for compatibility, but the value is no longer hardcoded in the filter.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/repository/JvmRepository.kt
tngusdl3719: Addressed

scavenger-api/build.gradle.kts
tngusdl3719: Addressed

scavenger-api/src/main/kotlin/com/navercorp/scavenger/repository/SnapshotNodeDao.kt
tngusdl3719: Addressed

scavenger-api/src/test/kotlin/com/navercorp/scavenger/mcp/ScavengerMcpToolsTest.kt
tngusdl3719: Addressed. `ScavengerMcpToolsTest` now deserializes the JSON response and asserts on structure/types instead of checking string fragments, while `ScavengerMcpServiceTest` continues to cover the deeper typed behavior.

tngusdl3719:
> **Additional review based on the intended design direction for this feature.**
> 
> We have design decisions for the MCP server that differ from this PR's approach. Here are the key points:
> 
> ### 1. Authentication: apiKey-based customer scoping (Critical)
> Scavenger agents already authenticate via `apiKey` (= `licenseKey` in the DB). Each apiKey is bound to a specific customer (workspace). The MCP server should use the same mechanism:
> 
> * Client passes apiKey (e.g. via `X-Scavenger-Api-Key` header)
> * Server calls `LicenseService.check(apiKey)` → resolves `customerId`
> * All subsequent tool calls are scoped to that `customerId` only
> * **`customerName` parameter should be removed from all tools** — customer is determined by the apiKey
> 
> Currently, the PR exposes all customers via `listCustomers()` with no authentication. Anyone with access to the `/mcp` endpoint can query any workspace's data.
> 
> ### 2. Read-only tools only — remove mutation tools (Major)
> `createSnapshot` and `refreshSnapshot` modify data. MCP tools should be **read-only** — snapshot management belongs in the web UI. Please remove these.
> 
> ### 3. Too many tools — focus on core scenarios (Major)
> The core MCP use cases are: (1) check if a method is used, (2) list dead code, (3) get usage summary, (4) check who calls a method, (5) list snapshots, (6) find newly dead code since deploy.
> 
> Tools that become unnecessary with apiKey auth: `listCustomers`, `resolveWorkspace`. Tools that overlap with the above: `searchMethods`, `getClassUsageSummary`, `getPackageUsageSummary`, `getSnapshotTree`. Tools that should not exist: `listApplications`, `listEnvironments` (not relevant to MCP use cases).
> 
> **Missing**: a `get_method_callers` tool using callstack data — this is the most valuable tool for safe deletion decisions, as it tells you whether removing a dead method will break any caller.
> 
> ### 4. Automatic snapshotId resolution (Major)
> Most tools currently require an explicit `snapshotId`, which forces the AI to call `listSnapshots` first on every interaction. When `snapshotId` is omitted, the server should automatically use the latest snapshot. This significantly reduces tool call overhead.
> 
> ### 5. Standardized error responses (Minor)
> Please define a consistent error format (e.g. `{ "error": "SNAPSHOT_NOT_FOUND", "message": "...", "hint": "..." }`) instead of mixing error state into `Map<String, Any?>` with `"error" to true`.

---

Thanks for clarifying the intended MCP direction. This is broader than a convention fix and would significantly re-scope the current PR. Before I rework it, should I reshape this PR around the apiKey-scoped read-only core toolset, or would you prefer to close this one and follow up with a new PR aligned to that design?

.claude/settings.local.json
```diff
@@ -0,0 +1,28 @@
+{
```
taeyeon-Kim, line 1: This is a local-only settings file containing hardcoded local paths (`/Users/suhyeon/workspace/scavenger/`) and personal permission grants. It should not be committed. Please remove this file from the PR and ensure `.claude/settings.local.json` is in `.gitignore`.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/mcp/ScavengerMcpService.kt
```diff
@@ -0,0 +1,312 @@
...
+import com.navercorp.scavenger.dto.ApplicationDetailDto
+import com.navercorp.scavenger.dto.CustomerDto
+import com.navercorp.scavenger.dto.EnvironmentDetailDto
+import com.navercorp.scavenger.dto.McpClassUsageSummaryDto
+import com.navercorp.scavenger.dto.McpCleanupCandidateDto
+import com.navercorp.scavenger.dto.McpMethodReferenceDto
+import com.navercorp.scavenger.dto.McpMethodUsageDto
+import com.navercorp.scavenger.dto.McpMethodsNotInvokedSinceLastDeployResultDto
+import com.navercorp.scavenger.dto.McpPackageUsageSummaryDto
+import com.navercorp.scavenger.dto.McpResolvedWorkspaceDto
```
taeyeon-Kim, line 12: `ScavengerMcpService` directly injects `SnapshotNodeDao`, `MethodRepository`, `GithubMappingRepository`, and `JvmRepository`, bypassing the service layer.

The existing pattern is `Controller → Service → DAO/Repository` (e.g. `SnapshotController` → `SnapshotNodeService` → `SnapshotNodeDao`).

The 3 new DAO methods (`findAllUnusedMethodNodes`, `countMethodUsageSummary`, `findMethodsNotInvokedSince`) should be exposed through `SnapshotNodeService` as delegate methods. Similarly for `MethodRepository` and `JvmRepository` — go through their respective services.

This keeps the layer architecture consistent and makes these methods reusable from both MCP and REST API layers.

scavenger-api/src/main/kotlin/com/navercorp/scavenger/dto/McpMethodReferenceDto.kt
```diff
@@ -0,0 +1,30 @@
...
+        fun from(entity: SnapshotNodeEntity): McpMethodReferenceDto {
+            return McpMethodReferenceDto(
+                signature = entity.signature,
+                className = entity.parent,
+                lastInvokedAtMillis = entity.lastInvokedAtMillis,
+                lastInvokedDate = entity.lastInvokedAtMillis?.toDateString()
+            )
+        }
+
+        private fun Long.toDateString(): String {
```
taeyeon-Kim, line 24: `Long.toDateString()` is duplicated in 4 places with the identical implementation:
- `McpMethodReferenceDto.kt` (here)
- `McpMethodUsageDto.kt`
- `McpSnapshotNodeDto.kt`
- `ScavengerMcpService.kt`

Please extract into a single shared location — e.g. a package-level extension function in the `mcp` package.

taeyeon-Kim:
Please address the inline review comments above along with the design direction from the previous review before proceeding further. Let's resolve the current state first.

taeyeon-Kim:
Thanks for kickstarting the MCP server work here 🙏

The design was reworked in #170 after two cross-model reviews, and has since been implemented in phases:

- #171 — authentication + response envelope
- #173 — read services + query indexes
- #177 — server transport, the five tools, and the response envelope (+ `doc/mcp-guide.md`)

Since #177 supersedes this prototype, closing this one. The history stays here for reference — thanks again for the initial exploration!

PR #168, opened 2026-04-14
Fix gRPC UNAVAILABLE error

NineTIN:
https://github.com/naver/scavenger/issues/78 와 동일한 `StatusRuntimeException: UNAVAILABLE` 문제를 겪고 있었고,
여러 가지로 확인해보다가 원인으로 보이는 부분을 수정해봤더니 문제가 재현되지 않아 PR 보냅니다.

## 환경
- 쿠버네티스 환경
- Scavenger Agent → [plaintext gRPC] → Ingress L7 (ALB, HTTPS)

## 확인한 사항
- HTTPS 통신은 서비스 <-> collector 간 문제 없이 동작
- gRPC는 reflection이 꺼져 있어서 proto를 직접 지정해 호출해보니 정상적으로 통신됨  
→ 네트워크나 단순 연결 문제는 아닌 것으로 판단했습니다.

## 원인
Ingress(L7)가 HTTPS(TLS) 연결을 기대하는 상태에서 Agent는 `usePlaintext()` 설정으로 평문 gRPC 요청을 보내고 있었습니다. 

https://github.com/naver/scavenger/blob/c0fcd76c9a819a504b5ad2052e170be7baa7f174/scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/publishing/GrpcClient.java#L100-L104

이 구성에서는 TLS Handshake가 성립되지 않아 연결이 즉시 종료되고 그로인해 Agent에서는 스트림이 비정상 종료된 것으로 인식되어 `UNAVAILABLE` 에러가 발생하는 것이 아닐까? 하고 가정했습니다.

## 해결
- Collector URL이 HTTPS인 경우(TLS 사용), gRPC 클라이언트가 `usePlaintext()` 대신 `useTransportSecurity()`를 사용하도록 조건 분기 추가

## 테스트
- 동일한 환경(Scavenger Agent → [plaintext gRPC] → Ingress L7 (ALB, HTTPS))에서 재현 확인
- 수정 적용 후, 동일 조건에서 `StatusRuntimeException: UNAVAILABLE` 발생하지 않음을 확인
- gRPC 스트림이 정상적으로 유지되는 것 확인

### DIFF
--- a/scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/publishing/GrpcClient.java
+++ b/scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/publishing/GrpcClient.java
@@ -32,11 +32,11 @@ public class GrpcClient implements AutoCloseable {
     private ManagedChannel channel;
     private GrpcAgentServiceGrpc.GrpcAgentServiceBlockingStub stub;
 
-    public static GrpcClient plaintext(String host) {
+    public static GrpcClient createPlaintext(String host) {
         return new GrpcClient(host, false);
     }
 
-    public static GrpcClient tls(String host) {
+    public static GrpcClient createTls(String host) {
         return new GrpcClient(host, true);
     }
 
--- a/scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/publishing/GrpcClient.java
+++ b/scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/publishing/GrpcClient.java
@@ -7,12 +7,11 @@ import java.util.logging.Level;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import com.navercorp.scavenger.model.CallStackDataPublication;
-
 import io.grpc.ManagedChannel;
 import io.grpc.okhttp.OkHttpChannelBuilder;
 import lombok.extern.java.Log;
 
+import com.navercorp.scavenger.model.CallStackDataPublication;
 import com.navercorp.scavenger.model.CodeBasePublication;
 import com.navercorp.scavenger.model.GetConfigRequest;
 import com.navercorp.scavenger.model.GetConfigResponse;
@@ -29,14 +28,14 @@ public class GrpcClient implements AutoCloseable {
     private static final Pattern DATA_SIZE_PATTERN = Pattern.compile("^([+\\-]?\\d+)([a-zA-Z]{0,2})$");
 
     private final String host;
-
+    private final boolean useTls;
     private ManagedChannel channel;
     private GrpcAgentServiceGrpc.GrpcAgentServiceBlockingStub stub;
 
-    public GrpcClient(String host) {
-        log.info("[scavenger] creating new grpc client. host is " + host);
+    public GrpcClient(String host, boolean useTls) {
+        log.info("[scavenger] creating new grpc client. host is " + host + " tls=" + useTls);
         this.host = host;
-
+        this.useTls = useTls;
         createNewChannelIfShutdown();
     }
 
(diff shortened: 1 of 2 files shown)


scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/publishing/GrpcClient.java
```diff
@@ -29,14 +28,14 @@ public class GrpcClient implements AutoCloseable {
...
 
-    public GrpcClient(String host) {
-        log.info("[scavenger] creating new grpc client. host is " + host);
+    public GrpcClient(String host, boolean useTls) {
+        log.info("[scavenger] creating new grpc client. host is " + host + " tls=" + useTls);
         this.host = host;
-
+        this.useTls = useTls;
         createNewChannelIfShutdown();
     }
```
taeyeon-Kim: 💭 **Nit — extracting named factories to hide the boolean flag** 🧹

The `boolean useTls` on the public constructor is a classic flag-argument smell: at call sites, `new GrpcClient(url, true)` doesn't tell the reader what `true` actually means. Since we only ever construct a client in one of two well-defined modes, named factory methods express the intent explicitly without any extra machinery:

```java
public static GrpcClient plaintext(String host) { return new GrpcClient(host, false); }
public static GrpcClient tls(String host)       { return new GrpcClient(host, true); }

private GrpcClient(String host, boolean useTls) {
    log.info("[scavenger] creating new grpc client. host=" + host + " tls=" + useTls);
    this.host = host;
    this.useTls = useTls;
    createNewChannelIfShutdown();
}
```

Then in `Publisher#getGrpcClient()`:

```java
grpcClient = useTls ? GrpcClient.tls(target) : GrpcClient.plaintext(target);
```

The `boolean` stays as a private field — no new enum or value object, so the agent stays lightweight — but the public surface becomes self-documenting. Feel free to skip if you'd rather keep the diff minimal. 🙂

taeyeon-Kim (commented):
🎉 The root-cause analysis is spot on.

The V5 init-config endpoint at `AgentController.kt:45` does `request.requestURL.split("/")[2]`, which drops the URL scheme entirely — so the agent has no way to decide whether the gRPC channel should use TLS. Compared to the V4 endpoint at `AgentController.kt:35` (`"${splitUrl[0]}//${splitUrl[2]}"`), which preserves the scheme, this was a clear design oversight.

Grounding the decision on `config.serverUrl`'s scheme is the only correct answer here: the collector only listens on plaintext HTTP via Armeria (`application.yml`, `server.port: -1` + `protocols: HTTP`), which means any `request.requestURL`-based detection would always be wrong behind a TLS-terminating proxy — exactly the environment this PR is fixing. 👍

Left one small nit about the boolean flag on the public constructor. Happy to approve once that's addressed (or feel free to decline if you'd rather keep the diff minimal). 🙏

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/publishing/GrpcClient.java
NineTIN: Thanks for taking a look at the PR.
I’ve updated it based on your comments.

taeyeon-Kim (approved):
👍 👍

kojandy (approved):
🔥

NineTIN:
머지된 이후 Maven Repository에는 언제쯤 반영되는지 알 수 있을까요?

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/publishing/GrpcClient.java
```diff
@@ -29,14 +28,22 @@ public class GrpcClient implements AutoCloseable {
...
     private final String host;
-
+    private final boolean useTls;
     private ManagedChannel channel;
     private GrpcAgentServiceGrpc.GrpcAgentServiceBlockingStub stub;
 
-    public GrpcClient(String host) {
-        log.info("[scavenger] creating new grpc client. host is " + host);
-        this.host = host;
+    public static GrpcClient plaintext(String host) {
```
sohyun-ku, line 35: I’d suggest renaming these to something like `createPlaintext` and `createTls`.
Using `create` makes it immediately clear that these are factory methods responsible for instantiation, rather than configuration or transformation.

sohyun-ku:
@NineTIN 
Thank you for the valuable PR👍
We are currently in the process of coordinating the overall project priorities and the next release cycle. 
Please understand that it may take some time before the actual release. We appreciate your patience in the meantime.

scavenger-agent-java/src/main/java/com/navercorp/scavenger/javaagent/publishing/GrpcClient.java
NineTIN: Updated the names

sohyun-ku (approved):

Issue #170, opened 2026-05-07
Add MCP server for AI-driven dead code analysis


taeyeon-Kim:
## Background

Modern development workflows are shifting toward AI coding agents (Claude Code, Cursor agent, Codex, Aider, etc.) that interact with codebases through tool calls. Scavenger's runtime dead-code data is uniquely valuable for these agents — it can ground AI reasoning in **actual production usage** rather than static heuristics.

PR #167 introduced an initial MCP server prototype but stalled. This issue proposes a redesigned MCP server that:
- Lets developers query Scavenger from any MCP-capable client with a **single config line**, no installation, no extra tooling
- Exposes a small set of well-described primitive tools that AI agents can compose into a wide variety of workflows
- Mirrors the existing Scavenger UI semantics so behavior is predictable

> 📜 **Revision note (2026-06-11).** This body was reworked to be the single authoritative design after two cross-model reviews (implementation audit + first-principles design review). The comments below are kept as the historical record of how the design evolved; where they differ from this body, **the body wins**.

## Goals

- Provide an MCP server endpoint embedded in `scavenger-api` over Streamable HTTP
- Authenticate via the existing `licenseKey` (header: `X-Scavenger-License-Key`)
- **Serve all tools via direct queries against `methods` / `invocations` / `call_stacks` — no materialized AI-only snapshot**
- Expose 5 read-style primitive tools that compose into a broad range of dead-code and impact-analysis use cases
- Provide rich `Use this tool when:` descriptions to maximize correct tool selection by AI agents

## 🎯 Correctness Principle

Scavenger's output feeds **code-deletion decisions**. A false "dead" verdict can delete live production code — so accuracy and explicit evidence limits take priority over response minimalism. Concretely:

- Empty results must distinguish "no data" / "feature disabled" / "outside instrumentation scope" — never let an agent read absence as proof of death
- Responses carry **collection-coverage metadata** (since when each (application, environment) has been reporting, agent liveness, per-method `knownSince`) — "no invocations in 30 days" is meaningless if collection started 5 days ago
- Tool descriptions frame results as **evidence, not deletion verdicts**
- No false caveats either (e.g., calling exhaustively-recorded data "sampled" makes an LLM hedge wrongly)

## Target User (P0)

**Developers using shell-based AI coding agents** (Claude Code, Cursor agent, Codex, etc.). Secondary users (Claude Desktop chat, etc.) get the same MCP endpoint at no extra cost.

[...]

PR #171, opened 2026-05-07
Add MCP authentication and envelope (#170 phase 1)

taeyeon-Kim:
## Summary

Phase 1 of #170 — **scope-reduced** after the 2026-05-20 redesign on #170 dropped the AI-managed snapshot model. This PR now lands only the foundation pieces that survive the redesign:

- License-key authentication for the `/mcp` endpoint
- Shared MCP response envelope and error model
- A small one-line fix to a long-standing AOP self-call bug uncovered during review

The AI-snapshot lifecycle (`AiSnapshotService`, async refresh, reserved-name handling, etc.) is **removed** from this PR. With the redesigned plan there is no `__ai__` snapshot at all — MCP tools will query `methods` / `invocations` / `call_stacks` directly. See [the redesign comment on #170](https://github.com/naver/scavenger/issues/170#issuecomment-4494193810) for the benchmark data and rationale.

### Authentication for `/mcp`
- `CustomerRepository.findByLicenseKey` (derived query)
- `ApiKeyAuthInterceptor` registered on `/mcp` and `/mcp/**`; resolves the `X-Scavenger-License-Key` header to a `customerId` and rejects requests with a sealed `McpError` body
- `mcp/McpAuthContext` — request-attribute key constants and `HttpServletRequest.requireCustomerId()` extension
- Cross-customer access attempts return a generic body (no existence leak)

### Shared response types
- `dto/McpError` (sealed code enum) and `dto/McpResponse<T>` envelope used by every MCP tool in later phases

### AOP self-call fix
- `@EnableAspectJAutoProxy(exposeProxy = true)` on `ScavengerApiApplication`
- Background: `SnapshotService.createSnapshot` already used `proxy(this).saveSnapshotWithLimit(...)` to invoke a `@Transactional(REQUIRES_NEW)` method on the same bean. Without `exposeProxy = true`, the `proxy(...)` helper's catch block silently returned the raw `this`, so the `REQUIRES_NEW` boundary was never applied. Adding the annotation makes the existing helper actually work — independent of MCP, and a fix the snapshot-count race relied on.
- Verified with a behavior-based test that runs the call inside an outer transaction set `rollbackOnly`. With the annotation, the inner write survives; without it (the prior production behavior), the inner write rolls back with the outer.

## Removed from this PR (compared to the original draft)

Per the #170 redesign:

- `AiSnapshotService`, `AiSnapshotRefresher`, `AiSnapshotState`
- `AsyncConfig` (`@EnableAsync` + MDC-aware executor)
- `McpProperties` (`staleThresholdMinutes` / `autoRefreshEnabled`) and the corresponding yml blocks
- `ReservedSnapshotNameException` and the `RESERVED_NAME_PREFIX` filters on `SnapshotService.listSnapshots` / `saveSnapshotWithLimit`
- The `__ai__` seed row (`test-data-set-1.1.5.sql`)

Auth + envelope is the only deliverable here.

## Test plan

- [x] `CustomerRepositoryTest.findByLicenseKey` — happy path + unknown key
- [x] `ApiKeyAuthInterceptorTest` (`@SpringBootTest` with seeded H2 data) — header missing / blank / invalid / valid
- [x] `SnapshotServiceTest` — REQUIRES_NEW survives outer rollback (proves the AOP fix)
- [x] `./gradlew :scavenger-api:test`
- [x] `./gradlew :scavenger-api:ktlintCheck`

Refs #170

### DIFF
--- a/scavenger-api/src/main/kotlin/com/navercorp/scavenger/ScavengerApiApplication.kt
+++ b/scavenger-api/src/main/kotlin/com/navercorp/scavenger/ScavengerApiApplication.kt
@@ -2,8 +2,10 @@ package com.navercorp.scavenger
 
 import org.springframework.boot.autoconfigure.SpringBootApplication
 import org.springframework.boot.runApplication
+import org.springframework.context.annotation.EnableAspectJAutoProxy
 
 @SpringBootApplication
+@EnableAspectJAutoProxy(exposeProxy = true)
 class ScavengerApiApplication
 
 fun main(args: Array<String>) {
--- a/scavenger-api/src/main/kotlin/com/navercorp/scavenger/config/WebConfig.kt
+++ b/scavenger-api/src/main/kotlin/com/navercorp/scavenger/config/WebConfig.kt
@@ -1,5 +1,6 @@
 package com.navercorp.scavenger.config
 
+import com.navercorp.scavenger.mcp.ApiKeyAuthInterceptor
 import org.slf4j.MDC
 import org.springframework.context.annotation.Configuration
 import org.springframework.core.Ordered
@@ -14,9 +15,13 @@ import jakarta.servlet.http.HttpServletResponse
 
 @EnableWebMvc
 @Configuration
-class WebConfig : WebMvcConfigurer {
+class WebConfig(
+    private val apiKeyAuthInterceptor: ApiKeyAuthInterceptor
+) : WebMvcConfigurer {
     override fun addInterceptors(registry: InterceptorRegistry) {
         registry.addInterceptor(MdcLoggingInterceptor())
+        registry.addInterceptor(apiKeyAuthInterceptor)
+            .addPathPatterns("/mcp", "/mcp/**")
     }
 
     override fun addResourceHandlers(registry: ResourceHandlerRegistry) {
--- a/scavenger-api/src/main/kotlin/com/navercorp/scavenger/dto/McpError.kt
+++ b/scavenger-api/src/main/kotlin/com/navercorp/scavenger/dto/McpError.kt
@@ -1,0 +1,18 @@
+package com.navercorp.scavenger.dto
+
+data class McpError(
+    val code: Code,
+    val message: String,
+    val hint: String? = null,
+    val retryable: Boolean = false
+) {
+    enum class Code {
+        AUTH_MISSING,
+        AUTH_INVALID,
+        METHOD_NOT_FOUND,
+        SNAPSHOT_NOT_READY,
+        INVALID_ARGUMENT,
+        REFRESH_FAILED,
+        INTERNAL_ERROR
+    }
+}
--- a/scavenger-api/src/main/kotlin/com/navercorp/scavenger/dto/McpResponse.kt
+++ b/scavenger-api/src/main/kotlin/com/navercorp/scavenger/dto/McpResponse.kt
@@ -1,0 +1,13 @@
+package com.navercorp.scavenger.dto
+
+data class McpResponse<out T>(
+    val ok: Boolean,
+    val data: T? = null,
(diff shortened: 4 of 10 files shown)


scavenger-api/src/main/kotlin/com/navercorp/scavenger/mcp/AiSnapshotService.kt
```diff
@@ -0,0 +1,105 @@
...
+            snapshot = snapshot,
+            ageMinutes = ageMinutesOf(snapshot),
+            refreshTriggered = true,
+            refreshError = if (refreshed == null) "refresh failed; previous snapshot retained" else null,
+        )
+    }
+
+    fun getOrCreate(customerId: Long): SnapshotEntity {
+        return findExisting(customerId) ?: proxy(this).createAiSnapshot(customerId)
+    }
```
sohyun-ku: In the newly added `getOrCreate`, is `proxy(this).createAiSnapshot(...)` actually going through the AOP proxy so that `@Transactiona`l gets applied? `AopContext.currentProxy()` works when `@EnableAspectJAutoProxy(exposeProxy = true)` is set, but I don't see it on ScavengerApiApplication — so I was wondering whether it might be falling back to returning the raw this instead.

taeyeon-Kim: Ah, this turned out to be a pre-existing bug — `SnapshotService.saveSnapshotWithLimit` was using the same `proxy(this)` pattern and was already silently broken (the helper's catch block in `Aops.kt` swallows the `IllegalStateException` and falls back to returning the raw `this`, so `@Transactional(REQUIRES_NEW)` was never applied).

Took this chance to add `@EnableAspectJAutoProxy(exposeProxy = true)` on `ScavengerApiApplication` so the helper actually goes through the AOP proxy. I also reinforced the tests with behavior-based verification — REQUIRES_NEW must commit independently of an outer rollback — which would have caught this earlier. Thanks for catching it!

taeyeon-Kim:
@sohyun-ku Heads up — I scope-reduced this PR after the redesign discussion on #170. The full rationale (with benchmarks) is in [the redesign comment on #170](https://github.com/naver/scavenger/issues/170#issuecomment-4494193810), but in short:

- The AI-managed `__ai__` snapshot model is dropped. Direct queries against `methods` / `invocations` / `call_stacks` win on every tool scenario with three added indexes, while keeping data always-fresh and preserving env/app dimensions that `snapshot_nodes`'s `max()` aggregation throws away.
- All snapshot lifecycle code (`AiSnapshotService`, `AiSnapshotRefresher`, `AsyncConfig`, `McpProperties` staleness fields, `ReservedSnapshotNameException`, the `__ai__` seed) is removed from this PR.
- What stays: the auth + envelope pieces that survive the redesign (`ApiKeyAuthInterceptor`, `CustomerRepository.findByLicenseKey`, `McpError` / `McpResponse` DTOs).

Your earlier review on the `proxy(this)` self-call also surfaced a pre-existing bug in `SnapshotService.saveSnapshotWithLimit` — the same pattern was silently broken there too because `@EnableAspectJAutoProxy(exposeProxy = true)` was never set, so the helper's catch block fell back to raw `this` and `@Transactional(REQUIRES_NEW)` never applied. That fix stays in this PR (one annotation + a behavior-based regression test in `SnapshotServiceTest`).

Title and body updated; the branch history is squashed to two commits (auth + envelope, then the AOP fix). Ready for re-review. Thanks again for catching the proxy issue — it ended up being the trigger for the whole redesign.

sohyun-ku (approved):

sohyun-ku (approved):

kojandy (approved):

PR #173, opened 2026-06-11
Add MCP read services and query indexes (Phase 2 of #170)

taeyeon-Kim:
## Summary

Phase 2 of #170 — read services + index migration for the MCP server, implemented per the locked design in the issue body (all cross-model audit dispositions included; single PR as agreed).

### Index migration (`schema-1.1.5.sql`, changeset `scavenger:9`)
- `ix_invocations_cust_sighash (customerId, signatureHash)`
- `ix_methods_cust_decltype (customerId, declaringType)`
- `ix_call_stacks_cust_sighash (customerId, signatureHash)`
- No `IF NOT EXISTS` (MySQL 8 doesn't support it on `CREATE INDEX`; H2 does — would mask the failure in tests). `--rollback` included. Shard-local, leading `customerId` — no vschema change.
- 🚦 **EXPLAIN gate — passed** (details below): real MySQL 8.0.46 + a real 2-shard Vitess cluster (vttestserver v25). The remaining environment-specific check is operational only: Online DDL duration of the three `CREATE INDEX` statements on the production `invocations` table.

### Read services (3)
| Service | Powers | Notes |
|---|---|---|
| `MethodUsageQueryService` | `is_method_used`, `get_pr_impact` | single shared `IN`-list query; "used" = INVOKED in **any** (app, env) via `MAX(CASE …)` aggregation, optional `env` scope; `get_pr_impact` explicitly returns `unknownSignatures` (absence must never read as "dead") |
| `StaleMethodSearchService` | `get_stale_methods` | corrected aggregation query (`GROUP BY` + `MAX/HAVING`), UI-parity filters (`garbage`/`modifiers`/`declaringType`), left-anchored escaped prefix, seek cursor on `signatureHash` |
| `MethodCallerQueryService` | `get_method_callers` | reuses the `CallStackSql` JOIN pattern (new SQL method); zero call-stack rows → typed `DISABLED_OR_NO_DATA` state instead of an empty list |

Method identification is **by signature string** — the API never computes signatureHash; it reads the stored hash from the matched `methods` row (avoids the `HashGenerator` `%x` leading-zero trap and a new module dependency).

### Envelope & contracts
- Envelope meta (`dataFreshness` / `coverage` / `pagination` / `notice` + their DTOs) is **deferred to Phase 3**, landing in the same PR that assembles it — defining it here had zero producers and silently added four null fields to Phase 1's auth error JSON. `McpResponse` stays `ok`/`data`/`error`
- Dead `McpError.Code` values `SNAPSHOT_NOT_READY` / `REFRESH_FAILED` removed
- Typed failure contract: services throw `McpException(McpError)`; Phase 3 maps to the envelope
- Input limits at service entry: signature ≤ 1000, signatures[] ≤ 200, prefix ≤ 256 (`declaringType` is `VARCHAR(256)`), limit ≤ 100

### Ops hardening
- MCP-only query protection: `@Transactional(readOnly = true, timeout = 10)` on the three services — UI query paths untouched
- Auth-path errors always answer with the JSON envelope (`INTERNAL_ERROR`, retryable), never Spring's default error page
- ~~licenseKey lookup cache~~ — considered, then **dropped as premature**: `/mcp` traffic is human-task frequency (unlike the collector's machine-frequency agent polling), the uncached path is a sub-ms `UNIQUE (licenseKey)` point lookup, and the cache would have added a 24h revocation lag plus two dependencies

### Env resolution (A1)
- `EnvironmentRepository.findByCustomerIdAndName` (nullable) + `McpEnvironmentResolver`: unknown env → `INVALID_ARGUMENT` (hint: use `list_scopes`), disabled env → `INVALID_ARGUMENT`; never collapsed into "0 results"

### Fixtures (`test-data-set-1.1.5.sql`, changeset `scavenger:10`)
- customer 1: `prod` (enabled) + `staging` (disabled) environments, cross-env divergence (`usedInTestOnly`: INVOKED in test / NOT_INVOKED in prod), never-invoked, abstract, and garbage methods
- customer 2: no call_stacks rows (powers the `DISABLED_OR_NO_DATA` state test + cross-customer isolation tests)
- `EnvironmentRepositoryTest.findAllByCustomerId` count updated 1 → 3 accordingly

## 🚦 EXPLAIN gate results

**MySQL 8.0.46 (benchmark scale: 50k methods × 60k invocations × 60k call_stacks, single customer, `ANALYZE TABLE` applied).** `schema-1.1.5.sql` applies cleanly — no `IF NOT EXISTS` syntax issue; all three indexes created.

| # | Shape | Plan | Actual |
|---|---|---|---|
| ① | usage (3-sig IN) | invocations via **`ix_invocations_cust_sighash`**; methods via customerId index + signature filter (the accepted by-signature shard-local scan) | 38.5 ms |
| ② | stale env+idleDays (workhorse) | join via **`ix_invocations_cust_sighash`** | 224 ms (issue H2 benchmark: 238 ms) |
| ③ | stale prefix LIKE (escaped, left-anchored) | **`ix_methods_cust_decltype` index range scan** | 0.02 ms |
| ④ | callers | **`ix_call_stacks_cust_sighash`** + existing `ix_methods_signaturehash_identity` single-row lookup | 0.07 ms |
| ⑤ | existsAnyCallStack | index-only | 0.005 ms |

**Vitess (vttestserver v25 snapshot, 2 shards, hash vindex on customerId — same layout as `vschema_shared-1.1.4.json`).**
- The three `CREATE INDEX` statements applied **through vtgate** (translated to `ALTER TABLE ADD KEY`, scattered to both shards) — migration DDL is Vitess-compatible.
- `VEXPLAIN PLAN` for all four query shapes (usage incl. LEFT JOIN + GROUP BY, stale workhorse incl. HAVING/ORDER/LIMIT, callers, exists): **`Route / EqualUnique / Vindex: hash`** — every query pushes down whole to a single shard. No scatter, no cross-shard work.
- In-shard `EXPLAIN` through vtgate confirms the same index choices as standalone MySQL: `ix_invocations_cust_sighash` (key_len 139), `ix_methods_cust_decltype` (range + ICP), `ix_call_stacks_cust_sighash` (key_len 138 with a matching hash).

Remaining operational item (not blocking correctness): measure Online DDL duration of the index build on the production-size `invocations` table during rollout.

## Test plan

- [x] `McpEnvironmentResolverTest` — name→id, unknown/disabled/blank, cross-customer
- [x] `MethodUsageQueryServiceTest` — invoked/never-invoked, **cross-app aggregation (B2 scenario)**, cross-env divergence, garbage/unknown → `METHOD_NOT_FOUND`, customer isolation, bulk found/unknown split, size/blank/oversize validation
- [x] `StaleMethodSearchServiceTest` — abstract/garbage/cross-app-invoked exclusion, env scoping, idleDays boundaries, **cursor paging without duplicates or loss + terminal null cursor**, literal `%`/`_` prefix, validation, isolation
- [x] `MethodCallerQueryServiceTest` — direct callers + last invocation, empty-with-data vs `DISABLED_OR_NO_DATA`, env scoping, unknown signature
- [x] `EnvironmentRepositoryTest`, `ApiKeyAuthInterceptorTest` (existing, behavior-based)
- [x] EXPLAIN gate: MySQL 8.0.46 + 2-shard Vitess (vttestserver) — see section above
- [x] Full `./gradlew test` green (collector fixture collision resolved by isolating new data under customer 2)
- [x] Self-review + 3 parallel review agents; accepted findings applied (shared signature validation, DbRow relocation to the api-side entity package, interceptor catch-all → JSON envelope, static not-found messages); license cache and the unused envelope-meta DTOs subsequently removed as premature

Refs #170


sohyun-ku (approved):
👍

kojandy (approved):

